PR #1, opened 2015-06-22
Allow testbase overriding in the config

grigorescu:
This branch allows for an alternative in the config file to set a different testbase.


rsmmr:
This has been merged.

PR #5, opened 2017-10-13
Fix TTY output test to work on Linux and BSD

ckreibich:
This specifically tests whether script supports the -c option (which
only exists on Linux) and uses this to distinguish BSD- and Linux-
specific implementations of the script invocation. The BSD
requirements test uses negation of the Linux one because running
script without -c on Linux causes it to enter a transcript session,
hanging btest.


ckreibich:
Fyi, I couldn't find a good way to just use one test script but distinguish BSD vs Linux, so I just made it two separate tests. Since the tests themselves are pretty simple, I feel this amount of duplication is okay, but feel free to let me know if you think otherwise.

The only actual discrepancy in the test implementations is how we invoke "script".

dnthayer:
In order to avoid the possibility of those two test files getting out-of-sync, I've
created an alternative fix that uses a wrapper script for the "script" command.
This avoids the need to split the test file (and perhaps the wrapper script can be
useful in the future for other tests).

rsmmr:
Fixed through alternative patch

PR #6, opened 2017-10-18
Tweak -A|--show-all to use only coloring, not cursor navigation

ckreibich:
Doing so lets us use colored output in contexts that don't deal well
with cursor navigation. The Console class now only uses color codes,
while the derived class CompactConsole also uses cursor placement to
overwrite output of succeeding tests.

Also a minor bit of renaming to convey that the _console* methods are
private, and to name function arguments a bit more consistently.

### DIFF
--- a/btest
+++ b/btest
@@ -1158,7 +1158,12 @@ class Standard(OutputHandler):
         self.output(test, msg)
 
 class Console(OutputHandler):
-    """Output handler that writes compact progress report to the console."""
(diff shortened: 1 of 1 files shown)


ckreibich:
Fyi folks, the context here is that in our btest executions in Gitlab we were still seeing line truncations in the build logs. In that setting, we use btest's output and place it in a surrounding log context (with timestamps etc), and would like to keep the colors test results since it helps quick navigation.

Thanks!

PR #8, opened 2018-05-23
Provide the name of the next upcoming test via TEST_NAME_NEXT env var

ckreibich:
This allows tests to adjust their behavior depending on upcoming
tests. For example, tests that would individually engage in expensive
set-up/tear-down may skip some of that work if they know that the next
test will just re-establish conditions that the current test would
tear down.

For the final test, TEST_NAME_NEXT remains unset.

Includes a bit of refactoring to separate the lock-guarded
identification of the next suitable test, which modifies
TestManager._tests, from the derivation of the actual test(s) to run,
and a minor bugfix in TestManager.threadRun(), where it looks like we
invoked saveTiming() repeatedly on the same tests (due to the use of
all_tests).

### DIFF
--- a/btest
+++ b/btest
@@ -324,71 +324,72 @@ class TestManager(multiprocessing.managers.SyncManager):
     def threadRun(self, thread_num):
         signal.signal(signal.SIGINT, signal.SIG_IGN)
 
-        all_tests = []
+        tests = self._produceTests(thread_num)
+
+        for test in tests:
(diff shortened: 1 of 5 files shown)


rsmmr:
Closing for now in anticipation of a revised version (discussed in person: in parallel mode, the patch may lead to individual expensive tests preventing subsequent ones from executing even if threads become available).

PR #9, opened 2018-11-05
topic/jsiwek/port-command

jsiwek:
Adds `TEST-PORT` command and `PortRange` option to facilitate parallelizing tests that require binding/listening on a TCP port.

### DIFF
--- a/README
+++ b/README
@@ -428,6 +428,11 @@ configuration file:
     measure the execution times of tests. By default, BTest searches
     for ``perf`` in ``PATH``.
 
+``PortRange``
+    Specifies a port range like "10000-11000" to use in conjuction with
+    ``@TEST-PORT`` commands.  Port assignments will be restricted to this
+    range.  The default range is "1024-65535".
+
 .. _environment variables:
 
 Environment Variables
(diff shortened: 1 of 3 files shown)


jsiwek:
What OS specifically ? (I've tested various Linux + macOS and Travis runs Linux + various Python versions without seeing this)

Is there anything at all in `.tmp/tests.ports/.tmp/ports/output` ?

Does the following command throw any exceptions on your test system?

```
python -c "import socket; sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM); sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 0); sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 0); sock.bind(('', 1024))"
```

The test does require the user to be able to bind at least 3 TCP ports in the 1024-65535 range.  If they can't, I'd expect the test to hang for a while: every 15 seconds there's a warning of the form `failed to obtain {0} ports for test {1}, will try {2} more times`, up to a max of 5 attempts.

rsmmr:
Found it, probably Python version thing. This line:

```
sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 0)
```

leads to ``failure 'module' object has no attribute 'SO_REUSEPORT'``. And that exception then let's the port availability check fail. Once I remove that line, it works fine.

PR #10, opened 2018-11-11
Improve handling of Ctrl-C with concurrent thread execution, and fix XML output

rsmmr:


btest
```diff
@@ -297,8 +315,11 @@ class TestManager(multiprocessing.managers.SyncManager):
                 t.join()
 
         else:
-            # No threads, just run all directly.
-            self.threadRun(0)
+            # No threads, just run one directly. We still spawn a subprocess
```
jsiwek: Maybe just set `num_threads` to 1 when `Options.thread` is zero and then collapse these branches into one.

rsmmr:
Note for myself: Check what exactly happens at CTRL-C time to currently running scripts. They should get a signal, too, but still run any cleanup traps.

rsmmr:
I have an updated version, need to test it a bit more.

rsmmr:
Tried this on Linux finally, and there's still some race condition there where, when hitting CTRL-C, btest is now not aborting. Pretty easy to reproduce with the btest test suite, `-j`, and hitting CTRL-C at some point while it runs.

rsmmr:
Closing this for now, needs more work.

PR #14, opened 2019-08-08
Add Test Retries

devbali:
Quoting @ckreibich as to what these changes accomplish:
> Adding a command line flag that lets you control the number of attempted retries (normally, 0)
> If > 0, running these retries whenever a test fails
> Preserving the output of each of the tests, so one can later go back and inspect them. You could devise some sort of automated directory naming, similar to how we add "-N", where N is a subtest, when a test currently consists of multiple tests. Perhaps "-retryN" or some such.
> A test that fails but later succeeds should be marked unstable — that's currently not a test status it supports.
> btest can output a JUnit XML file that summarizes the test results. It would be extra cool if in there too we can flag that a test is unstable.

Currently, I have -z and --retries as the option for retries, because -r is already occupied by the option to redo all failed tests once right now. I am assuming this new method is going to replace that, so retries can then occupy -r. Please give me the go ahead to do the replacement.


jsiwek:
"automatically retry a test failure up to N times within this test run" is different enough than "re-run any failed tests from previous run" that I'd like to see it as a separate option, not a replacement.

ckreibich:
Yeah, I agree as well. I should have called out the difference to `--rerun` — I think they're different use-cases. In particular, in a CI pipeline there's value in establishing that a given test is unstable, by retrying it right then and there, as opposed to making to overall btest invocations.

rsmmr:
Thanks, @devbali, this is nice. The one thing I'm wondering about is the new methods for pickling `Test`  instances. I get why they're necessary but I'm wondering if we could find a way to do this differently; it's a bit unfortunate to need this additional code, and makes it harder to maintain was well.

One idea: in `rerun()`,  could we call the test's `clone()` method to create a new copy instead of rescheduling the current one? `clone` would need a bit of tweaking to copy over the test's content as well; and not sure if there's more that would need to be re-initialized afterwards (`clone` is currently called only at initialization time). If this worked, we could then also reuse `clone`'s automatic numbering of tests with `-NN` instead of the separate `retry-NN` tags.

devbali:
Cloning makes sense, but I am not understanding how this will solve our need to have __getstate__ and __setstate__ functions as you would still have to add the test object with attributes of types not defined at the "top level" (in this case CmdLine Objects) to a shared list.
There's a couple of things I was thinking about to avoid __getstate__ and __setstate__:

- Remove the CmdLine Class, and use something like tuples instead. This is perhaps the easiest thing to do, and while it is not elegant at all, it works for removing the __getstate__ and __setstate__ functions from the Test class and just using its __dict__ for pickling. The __setstate__ function that I added basically does nothing more than just using this CmdLine __dicts__ instead of references. And this won't be a very significant change, considering this is the entire body of the CmdLine class:
```
 class CmdLine:
     def __init__(self, cmdline, expect_success, part, file):
         self.cmdline = cmdline
         self.expect_success = expect_success
         self.part = part
         self.file = file
 ```
-  We could also rerun on the Test level instead of the Test Manager level, so maybe a .run() on a Test Object would run again on the same thread if it fails and reruns are remaining

devbali:
Ignore my previous (now deleted comment)

Issue #15, opened 2019-08-25
Output timestamps on btest-progress stderr output


deejgregor:
This would help when looking into test failures in longer, more complex tests that use btest-progress and correlate with activities seen in log files.

deejgregor:
this issue was resolved when #16 was merged. I didn't need this separate issue, so closing.

PR #16, opened 2019-08-25
Output timestamps on btest-progress stderr output

deejgregor:
See #15.


ckreibich:
I like this one! :+1:

PR #17, opened 2019-10-01
If only new content is added optionally treat test as successful

szaydel:
This is just an idea I wanted to propose, and thought it is easier to just start with a PR. The thought here is that sometimes it may not matter if additions are made, but it does matter if something that was there is no longer there. It would be nice to have the option of saying: _if this is an addition to Baseline, ignore it and consider test a success, otherwise fail_. This approach can make some tests less prone to apparently failing as a result of a new feature, or some minor modification that adds something to output, which otherwise is effectively inconsequential. This also would be useful in cases where some ambiguity exists, like some contents are `grep`(ed) out of a log, or output, and while you know to expect certain results, there may be something else that only sometimes gets picked up, because of noise in the results being `grep`(ed), or unpredictable nature of logs, etc.

This PR lacks any documentation, etc., which I will happily complete if this will be seen as an acceptable addition.

### DIFF
--- a/btest-diff
+++ b/btest-diff
@@ -12,6 +12,7 @@
 #   TEST_DIFF_CANONIFIER
 #   TEST_DIFF_BRIEF
 #   TEST_DIFF_FILE_MAX_LINES
+#   TEST_DIFF_IGNORE_ADDED_LINES
 
 # Maximum number of lines to show from mismatching input file by default.
 MAX_LINES=100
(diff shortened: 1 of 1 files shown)


rsmmr:
Yeah, I see the utility and would generally be fine merging something like this. Two questions:

- How does this work with updating the baseline? In the current form the new baseline would now include those additions as well, which will then let the tests fail in environments where those aren't output.

- I'm wondering if there's a bit more of a broader solution where one could describe properties of the diff more generally: like "no deletions", "no additions", "at most five differences", etc. Not sure what the syntax for that would be, I imagine it would involve similar counting of `+`/`-` as in your patch, just be more flexible in  terms of conditions put on the results.

Maybe the answer to both is that `btest-diff` isn't quite the right place to implement this. We could add a separate tool that's more of a fuzzy matcher and knows to handle baselines in that case (which btw I have always wanted to do for Zeek logs: write a diff tool that's aware of the format and can be smarter about things like adding columns, changing UIDs, etc.)

szaydel:
Robin, thanks for a fast response! I had some of the same thoughts you had as well. Making this change in `btest-diff` was just too convenient and so I did not consider another tool, simply because it felt small enough, but having a more flexible tool makes a lot of sense, perhaps `btest-difff`, with third `f` standing for fuzzy. :)

Why I don't spend some time thinking about this, close this PR out for now and with your feedback do some experimentation in my spare time, and then can always come back with another PR.

rsmmr:
Ok, closing for now per your suggestion.

PR #22, opened 2020-04-27
Add support for emitting Chrome tracing logs

bbannier:
This patch add support for Chrome tracing logs. Logs can be converted to standalone HTML with e.g., `trace2html`. While this tool is primarily intended to profile Chrome itself, it is also an easy way to visualize e.g., the timing and scheduling of parallel processes, useful to identify BTests which contribute most to test execution wall time.

Below a rather uneventful example trace displayed in Firefox after processing with `trace2html`.
Screen Shot 2020-04-27 at 11 08 40 PM

### DIFF
--- a/README
+++ b/README
@@ -288,6 +288,12 @@ and 1 otherwise.
         the timing measurements are recorded as the new baseline instead
         of being compared to a previous baseline.
 
+    --trace-file=TRACEFILE
+        Record test execution timings in Chrome tracing format to the given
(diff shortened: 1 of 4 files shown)


bbannier:
> Can you add a test case? Also please update the README as well.

Done & done. The test got a little uglier than I wanted to (mainly due to different string handling between 2.7 and 3.x), but I hope it is still readable and useful.

PR #23, opened 2020-05-06
Terminate worker processes on keyboard interrupt.

bbannier:
Explicitly terminating the worker processes before cleaning up main
process prevents the workers from dying with a broken pipes to the
parent when trying to send back test results.

### DIFF
--- a/btest
+++ b/btest
@@ -313,13 +313,21 @@ class TestManager(multiprocessing.managers.SyncManager):
         if num_threads:
             threads = []
 
-            for i in range(num_threads):
-                t = multiprocessing.Process(name="#%d" % (i+1), target=lambda: self.threadRun(i))
-                t.start()
-                threads += [t]
+            try:
+                for i in range(num_threads):
+                    t = multiprocessing.Process(
+                            name="#%d" % (i+1),
+                            target=lambda: self.threadRun(i))
+                    t.start()
+                    threads += [t]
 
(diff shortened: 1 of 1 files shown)


bbannier:
This patch seems to work well for me on macos.

rsmmr:
Can you see if works on Linux too? I remember trying this a while ago and had trouble getting something to work on both platforms (but I don't recall the details, it's been a while).

bbannier:
> Can you see if it works on Linux too? I remember trying this a while ago and had trouble getting something to work on both platforms (but I don't recall the details, it's been a while).

I tested this on ubuntu-19.10. This patch does seem to preserve the existing behavior when running with `-j` and seems to correctly terminate workers on keyboard interrupt. The change here isn't much different from existing handling of `KeyboardInterrupt`.

rsmmr:
Tried out it out and what I'm seeing on both Mac and Linux is that ctrl-c is now terminating btest immediately, but keeps the current test processes running to completion in the background. As they finish up, the process list will eventually clean up.

That is a good improvement: the current behavior on ctrc-c also leaves current processes running, but dumps lots of stack backtraces to the console in addition. It's still not quite ideal yet: better would be to kill the child processes immediately as well, both to stop the CPU load but also in case one of them ends up hanging. 

I'll go ahead and merge, clearly better than before.

PR #30, opened 2020-10-12
Add new options -F/--abort-on-failure that will abort once at least o…

### LINKED ISSUE #29: Consider providing a flag to abort run once a single test has failed
While e.g., bisecting it might be more interesting to check whether the test suite passes at all than to see which particular tests have failed. Googletest provides a flag `--gtest_break_on_failure` which provides exactly that: following tests are skipped, and the test runner exits with a non-zero return value.

It would be great if btest would provide similar functionality.

Looking at the current implementation, the main difficulty in implementing this would be to ensure orderly shutdown of workers in a parallel run.

rsmmr:
…ne test has failed.

In parallel mode, we still let all currently executing tests run to
completion because it's tricky to kill them immediately. Might be
something for the future to address, probably by first getting rid of
non-threaded mode altogether to unify the logic.

Closes #29.

### DIFF
--- a/btest
+++ b/btest
@@ -2555,7 +2555,8 @@ try:
 
 except Abort as e:
     output_handler.finished()
-    print(e.message, file=sys.stderr)
+    print(e, file=sys.stderr)
+    sys.stderr.flush()
     os._exit(1)
 
 except (KeyboardInterrupt, IOError): # Ctrl-C can lead to broken pipe.


rsmmr:
@bbannier see if this works for you

bbannier (approved):
Thanks @rsmmr, this gets the job done for me.

I am not sure what is up with CI.

I personally now know of two places which would need to be touched when proper worker cleanup on abort gets fixed (one here, the other in the CTRL-C handling). It would be great to capture that in a ticket.

rsmmr:
Closing, it's already merged.

PR #33, opened 2020-11-05
Add support for multiple baseline directories

rsmmr:
This adds support for specifying a colon-separated list of baseline directories. `btest-diff` will search them in order when looking for a test baseline and comare against the first one it finds. When updating a baseline, `btest-diff` will always write the new version into the first directory. The baseline directories can be set in two ways:

1. Through the existing `BaselineDir` option in `btest.cfg`.
2. Through an environment variable `BTEST_BASELINE_DIR`. This also works inside an alternative's environment section.

Closes #33


rsmmr:
@vpax  see if this does what you were looking for

vpax:
@rsmmr works fine & just what I need, thanks!

rsmmr:
@ckreibich mind reviewing? just looking for a another pair of eyes as a double check

ckreibich:
Apologies for the delay here. I'll be looking this over tomorrow.

testing/tests/alternatives-baseline-dir.test
```diff
@@ -0,0 +1,19 @@
+# %TEST-EXEC-FAIL: btest -a baseline %INPUT
```
ckreibich, line 1: For all of these tests it would be helpful to add a brief comment to explain the rationale of the test. Our btests are generally pretty unapproachable because their name is usually their only indication of that rationale.

btest-diff
```diff
@@ -109,13 +112,19 @@ delete_tmps() {
 
 trap delete_tmps 0
 
+# First first available baseline across directories.
```
ckreibich, line 115: "First first"

btest
```diff
@@ -2529,7 +2541,9 @@ if Options.doc:
     outputDocumentation(tests, Options.doc)
     sys.exit(0)
 
-mkdir(BaselineDir)
+for d in BaselineDirs:
+    mkdir(d)
+
```
ckreibich, line 2546: Trailing whitespace.
```suggestion

```

btest
```diff
@@ -2301,26 +2307,32 @@ if Options.alternatives:
...
+    # If multiple alternatives set it, we pick the value from the first.
+    def get_env_from_alternative(env, opt, default, transform = None):
+        if env in os.environ: 
+            return
+
         for tag in Options.alternatives:
-            base = getOption("BTEST_TEST_BASE", None, section="environment-%s" % tag)
-            if base is None:
-                continue
+            value = getOption(env, None, section="environment-%s" % tag)
```
ckreibich, line 2319: Trailing whitespace.
```suggestion
            value = getOption(env, None, section="environment-%s" % tag)
```

btest
```diff
@@ -2301,26 +2307,32 @@ if Options.alternatives:
...
-    # variable, we check to see if we're using an alternative that sets
-    # it. We'll use the TestBase from the first matching alternative that
-    # does so.
-    if "BTEST_TEST_BASE" not in os.environ:
+    # Helper function that, if an option wasn't explicitly specified as an
+    # environment variable, checks if an alternative sets its through
+    # its own environment section. If so, we make that value our new default.
+    # If multiple alternatives set it, we pick the value from the first.
+    def get_env_from_alternative(env, opt, default, transform = None):
+        if env in os.environ:
```
ckreibich, line 2315: Trailing whitespace.
```suggestion
        if env in os.environ:
```

btest
```diff
@@ -2301,26 +2307,32 @@ if Options.alternatives:
...
-    # If TestBase wasn't explicitly specified as an environment
-    # variable, we check to see if we're using an alternative that sets
-    # it. We'll use the TestBase from the first matching alternative that
-    # does so.
-    if "BTEST_TEST_BASE" not in os.environ:
+    # Helper function that, if an option wasn't explicitly specified as an
+    # environment variable, checks if an alternative sets its through
+    # its own environment section. If so, we make that value our new default.
+    # If multiple alternatives set it, we pick the value from the first.
+    def get_env_from_alternative(env, opt, default, transform = None):
```
ckreibich, line 2314: Total nit if you want to be pylint-friendly: `transform=None` is more Pythonic.
```suggestion
    def get_env_from_alternative(env, opt, default, transform=None):
```

README
```diff
@@ -382,10 +382,21 @@ The following options can be set in the ``btest`` section of the
...
-    (note that the actual baseline files will be in test-specific
-    subdirectories of this directory).  By default, this is set
-    to ``%(testbase)s/Baseline``.
+    One or more directories where to store the baseline files for
+    ``btest-diff`` (note that the actual baseline files will be placed
+    into test-specific subdirectories of this directory).  By default,
+    this is set to ``%(testbase)s/Baseline``.
+
+    If multiple directories are to be used, they must be separated by
+    colons. ``btest-diff`` will then search them in order when looking
```
ckreibich, line 391: It might help here to explain what is searches for. It is presence of a baseline directory, but it could also be correctness of the baseline content, for example.

ckreibich (approved):
Good stuff — I think this addresses Vern's request, and then some. :-) I suggested some really minor nits. My only high-level comment is that I feel the code is actually fairly tricky, and I'm wondering whether we could get away with a bit less flexibility. Two thoughts here:

- Could the naming of the baselines be implicitly defined by the alternatives? I.e., if I'm in `[environment-foobar]`, `Baseline/foobar` would be implied?

- Will the alternatives mechanism ever really imply the use of more than one additional baseline? I.e., do we actually really need an n-element list, or could it just be a pair?

vpax:
> * Will the alternatives mechanism ever really imply the use of more than one additional baseline? I.e., do we actually really need an n-element list, or could it just be a pair?

I'm picturing that having more than one could be helpful for large changes the have multiple elements/stages.  For script optimization, for example, when turning on compilation-to-ZAM-including-inlining you might want to first look in `Baseline.ZAM/` and then in `Baseline.inlining/` (because inlining is stand-alone, not requiring compilation), so that the former doesn't have to include copies of stuff in the latter.  That said, it could turn out that the duplication is minor, or that the overlapping baselines are slightly different anyway, in which case there's no win from having the increased flexibility.  I'm not sure at this point whether for my particular use that will be the case.

testing/tests/alternatives-baseline-dir.test
rsmmr: Ack, point taken. Adding doc strings to the new tests, we even have `TEST-DOC` of course ..

rsmmr:
Thanks for reviewing!

> * Could the naming of the baselines be implicitly defined by the alternatives? I.e., if I'm in `[environment-foobar]`, `Baseline/foobar` would be implied?

The current scheme leaves a bit more freedom in where to put them, which could come in handy; and I think also it's also easier to follow by making the location explicit, rather than using a built-in convention. 

> * Will the alternatives mechanism ever really imply the use of more than one additional baseline? I.e., do we actually really need an n-element list, or could it just be a pair?

Per Vern's comment: maybe, maybe not. But even if we don't need it immediately, I'm not sure using a pair would actually make the code much easier, a list feels pretty natural to me here either way.

PR #37, opened 2020-12-03
Fix --abort-on-failure for expected failures.

### LINKED ISSUE #36: Option '--abort-on-failure' does not ignore known failures
7d34baef867aab0496e4adbe3c8e76abc0abb97e introduced the `--abort-on-failure` option which aborts processing as soon as the first failing test is encountered. Currently it has no special handling for known failures (tests marked with `@TEST-KNOWN-FAILURE`) and will abort processing whenever they are encountered, i.e., they are treated as true failures. This seems counterintuitive as globally BTest would not count such tests as failures in that even with such tests failing the global status could still be success.

rsmmr:
We no longer abort if there are only expected failures.

Closes #36.


testing/tests/abort-on-failure-with-only-known-fails.btest
```diff
@@ -0,0 +1,17 @@
...
+# %TEST-END-FILE
+
+# %TEST-START-FILE test2
+@TEST-EXEC: exit 1
+@TEST-KNOWN-FAILURE: This test is expected to fail.
+# %TEST-END-FILE
+
+# %TEST-START-FILE test3
+@TEST-EXEC: exit 0
+# %TEST-END-FILE
```
bbannier, line 17: This doesn't explicitly excercise `-F`. Maybe we should instead change this to execute `exit 1`, and add a fourth test which would succeed, but isn't run because of `-F`.

bbannier (approved):
Thanks for the patch @rsmmr, this appears to work for my setup.

PR #38, opened 2020-12-03
Default to "-j 1" behavior to make CTRL-C work on Linux, and avoid leaking btest processes

### LINKED ISSUE #35: CTRL-C of an unparallelized btest run doesn't work
On a "normal", unparallelized btest run (one without -j) nothing happens when I hit CTRL-C. I'm attaching a really small example to demonstrate. It contains a single test that just sleeps for 10 seconds, and that test just keeps going:

```
$ unzip testsuite.zip
$ cd testsuite
$ btest
[  0%] tests.test1 ...^C^C^C^C^C^C^C^C^C
```
Eventually the 10 seconds are up and I get

```all 1 tests successful```

I see this also with the Zeek testsuite, for example. The root cause seems to be that we block `SIGINT` in `TestManager.threadRun()`, and in unparallelized operation `threadRun()` doesn't execute in a `multiprocessing` child process, but the main one. I looked at a few options for fixing this, and everything I try ends up being ugly and adding a bunch of complexity. Essentially, once I remove the ignoring of `SIGINT`, handling `KeyboardInterrupt` either happens "too low" in the stack, meaning we just skip the current test and continue with the next (I'm not seeing a clean way to signal exit to the manager, other than abuse of `Options.abort_on_failure`?) or "too high", meaning multiprocessing stuff doesn't get cleaned up properly (I'm seeing the subprocess created in `runSubprocess()` continuing, even though its child properly exits).

Now ... the good news is that when I run btest with `-j 1`, i.e. not parallelized but multi-processed, everything works fine. CTRL-C makes btest exit cleanly (I get the cursor back, a correct summary message regarding tests that have completed at this point, the various processes/threads clean themselves up, etc). So I am wondering whether instead of contorting ourselves to support unparallelized operation, we should just default to j = 1 instead of j = 0, and remove a bit of code. Or perhaps I'm just not seeing the right fix.

Also, I believe @rsmmr observed that CTRL-C actually does work for him, and I'm now quite curious why. :-)

I'm on Linux, Python 3.9, fwiw.

[testsuite.zip](https://github.com/zeek/btest/files/5599716/testsuite.zip)

### PR DESCRIPTION
**Pull request by ckreibich**
See commit messages for details ... you were right @rsmmr, this is really fiddly stuff. :-) I wasn't able to get rid of the entirety of the non-multiprocessed code since interactive updates just behave more cleanly when running directly. That said, I like how things work now, it seems pretty consistent on FreeBSD and Linux.

I found another bug while playing with this: `--abort-on-failure` lead to the sync manager not getting terminated (on both Linux and FreeBSD). I think garbage collection and `os._exit()` don't mix well.

All in all I think this still isn't ideal, but it seems to behave more consistently now for me. If you'd like to test, I'd suggest running the testsuite normally, with `-j`, and with trying an interactive baseline update to confirm it works.

Fixes #35

### CODE REVIEW

File: btest [hunk 1]
```diff
@@ -1238,23 +1238,16 @@ class OutputHandler:
...
-
-        else:
-            try:
-                self._buffered_output[test.name] += [(msg, nl, file)]
-            except KeyError:
-                self._buffered_output[test.name] = [(msg, nl, file)]
+        try:
+            self._buffered_output[test.name] += [(msg, nl, file)]
+        except KeyError:
+            self._buffered_output[test.name] = [(msg, nl, file)]
```
bbannier, line 1250: Is there a reason we cannot take the old approach of only buffering if multithreaded which would only require a small change here?

```.diff
-        if not self.options().threads:
+        if self.options().threads < 2:
```

This would have the benefit of leaving a way to observe progress which is handy for long-running tests.

btest
```diff
@@ -2136,20 +2129,20 @@ def readTestFile(filename):
...
 
-        setattr(parser.values, option.dest, val)
+    if parser.rargs and not parser.rargs[0].startswith('-'):
+        try:
+            # Next argument should be the non-negative number of threads.
+            # Turn 0 into 1, for backward compatibility.
+            val = max(1, int(parser.rargs[0]))
+            parser.rargs.pop(0)
+        except ValueError:
+            error("If given an argument, -j|--jobs requires a positive integer")
```
bbannier, line 2142: We can leave the error reporting to `optparse`,

```.diff
-              error("If given an argument, -j|--jobs requires a positive integer")
+              raise optparse.OptionValueError("If given an argument, -j|--jobs requires a positive integer")
```

btest
```diff
@@ -2136,20 +2129,20 @@ def readTestFile(filename):
...
+            # Not reached
```
bbannier, line 2143: Leftover debugging comment?

btest
```diff
@@ -2259,6 +2252,11 @@ optparser.add_option("-F", "--abort-on-failure", action="store_true", dest="abor
 optparser.set_defaults(mode="TEST")
 (Options, args) = optparser.parse_args()
 
+# Update-interactive mode implies single-threaded operation
```
bbannier, line 2255: This is less a change required by your patch and more a fix for an existing bug 👏

testing/tests/threads.test
```diff
@@ -1,11 +1,26 @@
...
+#
+# 5 tests with 5 threads parallelize completely, but tests 4 and 5 are
+# serialized, so must end up in a single thread. The following groups
+# all test numbers run by a given thread on a single line, and confirms
+# that one of them includes 4, then 5.
 # %TEST-EXEC: btest -j 5 t1 t2 t3 t4 t5 2>&1 | ./normalize-output | grep "4.*5" | sed "s/[0-36-9] //" >output.j5
-# %TEST-EXEC: btest -j 1 t1 t2 t3 t4 t5 2>&1 | ./normalize-output >output.j1
+#
+# Single-thread operation is the default, so "-j 1" should yield same output.
+# %TEST-EXEC: btest -j 1 t1 t2 t3 t4 t5 2>&1 | cat >output.j1
```
bbannier, line 13: Since you are added another 5 seconds to this already slow test, could you remove the `sleep 1` from the test inputs to reduce the runtime?

bbannier (commented):
Really looking forward to this change, left a couple small nits.

btest
ckreibich: I thought of this too and was afraid I'd open another can of worms. It might well work. Have you tried?

testing/tests/threads.test
ckreibich: I don't think I added 5 seconds? The runtime is the same here. We could make those 5-stage steps 4-step ones. :-)

btest
```diff
@@ -2559,15 +2559,14 @@ try:
...
+    print(str(exc) or "Aborted with %s." % exc.__class__.__name__, file=sys.stderr)
     sys.stderr.flush()
-    os._exit(1)
-
-except (KeyboardInterrupt, IOError): # Ctrl-C can lead to broken pipe.
-    output_handler.finished()
-    print("Aborted.", file=sys.stderr)
+    # Explicitly shut down sync manager to avoid leaking manager
+    # processes, particularly with --abort-on-failure:
+    mgr.shutdown()
```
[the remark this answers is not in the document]
ckreibich: Good catch, yeah. I just checked the constructor code for the base classes and as far as I can tell it's safe to move that out of the try-block. I'll push an update shortly that adds this.

btest
```diff
@@ -2559,15 +2559,14 @@ try:
     total = succeeded + failed + skipped
     output_handler.finished()
 
-except Abort as e:
+# Ctrl-C can lead to broken pipe (e.g. FreeBSD), so include IOError here:
+except (Abort, KeyboardInterrupt, IOError) as exc:
     output_handler.finished()
-    print(e, file=sys.stderr)
+    print(str(exc) or "Aborted with %s." % exc.__class__.__name__, file=sys.stderr)
```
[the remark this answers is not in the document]
ckreibich: Nice, I like that. The `.__class__.__name__` thing is muscle memory for me.

btest
ckreibich: If you make these things suggestions in the future, Benjamin, I'm happy to add them with a click! That way you also show up in the commits.

ckreibich:
Okay @bbannier I've pushed all suggestions except for the buffering-avoidance one — simply because I feel we should try first. Thanks!

btest
bbannier: Yes, I have tried this and it seems to work. It also subjectively makes sense to me, but I might be missing something.

testing/tests/threads.test
bbannier: Ah right, missed that this was already there. Would still be great to get rid of the sleeps as the test doesn't seem to rely on them and they make this test take forever (the other slow one is `tests.groups`).

btest
```diff
@@ -2136,20 +2129,19 @@ def readTestFile(filename):
...
 
-        setattr(parser.values, option.dest, val)
+    if parser.rargs and not parser.rargs[0].startswith('-'):
+        try:
+            # Next argument should be the non-negative number of threads.
+            # Turn 0 into 1, for backward compatibility.
+            val = max(1, int(parser.rargs[0]))
+            parser.rargs.pop(0)
+        except ValueError:
+            raise optparse.OptionValueError("if given an argument, -j|--jobs requires a positive integer")
```
bbannier, line 2142: I have played a little with this, and found that triggering an exception here causes a regression. Previously this worked:

```.sh
$ btest -j test.threads
```

With this change we now need to be able to interpret non-flag tokens following `-j` as a number, and do not inject the default anymore:

```.sh
$ btest -j tests.threads
Usage: btest [options] <directories>

btest: error: if given an argument, -j|--jobs requires a positive integer
```

It would probably be best to keep the old behavior of assuming the default value if whatever follows `-j` cannot be parsed as a number, and let upstream machinery deal with any flag inconsistencies.

bbannier (commented):
I have been successfully using this today, and it appears to work fine.

I found one small issue in the change parsing of `-j`.

btest
ckreibich: Oh yeah definitely ... it wasn't my intention to break that. The testsuite should probably verify this, too.

We're basically trying to address a variant of the behavior that optparse explicitly doesn't support:

> Lots of people want an “optional option arguments” feature, meaning that some options will take an argument if they see it, and won’t if they don’t. This is somewhat controversial, because it makes parsing ambiguous: if -a takes an optional argument and -b is another option entirely, how do we interpret -ab? Because of this ambiguity, optparse does not support this feature.

I'll make it fall back to the all-CPUs default. And definitely thumbs up to migrating to argparse soonish...

btest
ckreibich: Agree, I just don't trust my intuition here any more. :-) Let's put in in then, thank you!

ckreibich:
Okay @bbannier, I hope this addresses everything.

Threading-test speedup:
```
$ time ../btest tests/threads.test 
all 1 tests successful

real    0m2.424s
user    0m1.058s
```
vs master:
```
$ time ../btest ./tests/threads.test
all 1 tests successful

real    0m14.390s
user    0m1.512s
sys     0m0.996s
```
I'm surprised it still takes 2s given what's left. I suspect that's more about the multiprocessing infrastructure setting itself up now than the test itself. Might be worth exploring a bit at a later point.

testing/Baseline/tests.exit-codes/out2
```diff
@@ -1,6 +1,7 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 t1 ... failed
 t2 ... failed
+2 of 2 tests failed
```
bbannier, line 4: TBH, I have no idea why this appeared now (even more so since the command executes three tests).

bbannier (approved):
LGTM, thanks @ckreibich.

testing/Baseline/tests.exit-codes/out2
rsmmr: The output probably comes now because the code does an extra flush at the end. It's just two tests because the 2nd exits with code 200 ("If a command returns the special exit code 200, the test is considered failed and all further tests are aborted.")

rsmmr:
Nice, merged! One additional observation: When I CTRL-C on macOS, I now always get `[Errno 32] Broken pipe`, which isn't that pretty. That's from here: 

```
print(str(exc) or "Aborted with %s." % type(exc).__name__, file=sys.stderr)
```

Should we split this `except` into two: (1) for abort and interrupt, and (2)  for I/O error; and then print different messages?

ckreibich:
Sure thing Robin, as you prefer! I love that message because it took me far too long to understand that it was the main reason for the different behaviors. ;-)

PR #39, opened 2020-12-05
Add "binary mode" to btest-diff

ckreibich:
This adds the binary-data support to `btest-diff` that @0xxon suggested in zeek/zeek#1309. Johanna, mind taking a look?

While digging through this I noticed the `"$@"` in the existing diff invocation ... so we're technically allowing additional command lines arguments to be passed through to diff. The new `--binary` flag shouldn't conflict here.


testing/tests/binary-mode.test
```diff
@@ -0,0 +1,38 @@
+# This tests btest-diff's binary mode. In binary mode we treat test
```
rsmmr, line 1: Nice test file!

PR #42, opened 2021-01-08
Apply TEST-REQUIRES to tests duplicated with TEST-START-NEXT.

rsmmr:
This was a bug that didn't transfer the requirement over to the clones.


bbannier (approved):
Can you rebase so the LGTM CI run can trigger?

rsmmr:
Done, but it says: "LGTM.com did not detect changes to any Python files "

bbannier:
> Done, but it says: "LGTM.com did not detect changes to any Python files "

It probably doesn't detect `btest` as a Python file without extra configuration.

PR #44, opened 2021-01-14
Support BTest installation via CMake

ckreibich:
This adds a CMakeList.txt that causes btest to be installed via CMake when bundled with Zeek. Since btest has a lot more scripts than zkg (where our CMake approach basically directly installs zkg and its module) I didn't want to list all of the scripts in the CMakeLists.txt. I instead tried a few ways of relying on a Python invocation of `setup.py`, since that approach means you can ignore the CMake side of things in the future. I tried two variants:

* Copy the whole source tree into the build directory, and allow `setup.py` to create any intermediate directories it requires, while controlling the scripts and module contents as needed via command-line options. This is relatively straightforward and creation of local `build` or `dist` directories is no problem.
* Run `setup.py` from the source directory, but ensure that it creates any intermediate directories (such as `build`) in the CMake build directory. This also works but it's more complex — you need to explicitly invoke `setup.py build` with a bunch of flags that control scripts and module, and then run it again so it installs based on those intermediate build directories.

In the end I found the former approach more straightforward, so that's what I went for. There's a small hiccup around the copying of the source tree: when your CMake build directory resides in the source tree (as it intutively would be with the usual `mkdir build && cd build && cmake ..` routine), CMake isn't smart enough to notice the recursive copying and blows up after a while. I addressed this by explicitly skipping the build directory, if contained in the source tree. This doesn't matter for the Zeek-bundled installation, where this constellation doesn't come up.

If we're okay with this approach, we could eventually use the same in zkg.

I've verified that the bundled btest installation works, and that a Zeek package installation via zkg works correctly (including test cases) even when the Zeek installation directory (with zeek, zkg, btest) isn't in PATH.

Resolves zeek/zeek#1360, companion PR to zeek/zeek#1366.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,38 @@
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+project(btest)
+
+# Install scripts
+install(DIRECTORY DESTINATION bin)
+install(PROGRAMS
+    btest
+    btest-ask-update
+    btest-bg-run
+    btest-bg-run-helper
+    btest-bg-wait
+    btest-diff
+    btest-setsid
+    btest-progress
+    sphinx/btest-diff-rst
+    sphinx/btest-rst-cmd
+    sphinx/btest-rst-include
+    sphinx/btest-rst-pipe
+    DESTINATION bin)
+
+if ( NOT PY_MOD_INSTALL_DIR )
+    # This is not a Zeek-bundled install, from which we'd inherit that
+    # variable.  Default to "home"-style install.
+    set(PY_MOD_INSTALL_DIR lib/python)
+endif ()
+
+# Install the Python module
+install(DIRECTORY DESTINATION ${PY_MOD_INSTALL_DIR})
+install(FILES sphinx/btest-sphinx.py DESTINATION ${PY_MOD_INSTALL_DIR})
+
+message(
+    "\n====================|  BTest Build Summary  |==================="
+    "\n"
+    "\nInstall prefix:      ${CMAKE_INSTALL_PREFIX}"
+    "\nPython module path:  ${PY_MOD_INSTALL_DIR}"
+    "\n"
+    "\n================================================================\n"
+)
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,8 @@
 
 from distutils.core import setup, Extension
 
+# When making changes to the following list, remember to keep
+# CMakeLists.txt in sync.
 scripts = [
     "btest",
     "btest-ask-update",


bbannier:
What's the use case for installing developer-only tooling such as BTest with Zeek when BTest can already today be installed and managed with e.g., `pip`? While bundling a pinned BTest version with Zeek seems reasonable, I have trouble seeing a use case for also installing it, e.g., we use BTest in Spicy as well, and if Zeek pushed its version on us via installation (Zeek comes from Homebrew for me), I could easily see this causing conflicts.

I think at the very least this should be behind an option (which probably should default to `OFF`).

bbannier:
What's the use case for installing developer-only tooling such as BTest with Zeek when BTest can already today be installed and managed with e.g., `pip`? While bundling a pinned BTest version with Zeek seems reasonable, I have trouble seeing a use case for also installing it, e.g., we use BTest in Spicy as well, and if Zeek pushed its version on us via installation (Zeek comes from Homebrew for me), I could easily see this causing conflicts.

ckreibich:
The main reason is that zkg tests packages upon install. Those tests usually require btest. If we keep things as-is, people get zkg out-of-the-box when installing Zeek from source, but package installations (for testable packages) will then fail unless they override the resulting warning or use `--skiptests`.

Zeek's own testing invokes btest from the submodule located in its own tree. Spicy could do the same if it wants to be certain about the version it uses. You can also express a btest version requirement in btest.cfg.

jsiwek:
On my first try, I got this:

```
  CMake Error at auxil/btest/CMakeLists.txt:19 (file):
  file COPY cannot copy file
  "/home/jon/pro/zeek/zeek/auxil/btest/testing/.tmp/btest-socket-740139" to
  "/home/jon/pro/zeek/zeek/build-rel/auxil/btest/btest/testing/.tmp/btest-socket-740139":
  No such device or address.
```

That's maybe not a common situation (that file exists while running btest's test suite, not sure why it was leftover) and I can easily workaround (just `rm` stuff), but...

> I didn't want to list all of the scripts in the CMakeLists.txt

Conceptually, might still be doing the same thing, but in different way: rather than maintain an inclusion-list, it's now maintaining an exclusion-list for the source-tree copy (e.g. beware if there's a `build/` dir, and possibly also other `.tmp` files per above problem, etc.).  A difference is that it's harder to verify the correctness of later while my imagination about what can go wrong with the inclusion-list approach is limited to "forgetting to list a file" and that's easy to understand and fix.

@ckreibich I have less concern about the approach of listing all the files to install, were there other ways of thinking that can go wrong?

ckreibich:
> On my first try, I got this

Argh, that's unfortunate. I didn't hit this.

> @ckreibich I have less concern about the approach of listing all the files to install, were there other ways of thinking that can go wrong?

Not immediately, no — you could list it all and install it explicitly. Since we already override some Python-internal logic (instead of trusting Python to figure out an installation path for the sphinx module we specify the new Zeek Python folder, for example), it's currently not tricky, and the set of files doesn't change much. An actual breaking point for this approach would occur if btest came with a module that would need compiling or some other form of intermediate representation for installation — basically, when Python-internal stuff starts to matter. It seems fairly unlikely that btest is headed for that level of complexity. I just felt it would have been nice to treat Python like a black box, but you've found other drawbacks with that.

ckreibich:
@jsiwek I've updated this to list files explicitly.

PR #53, opened 2021-08-18
Add Github action exercising the test suite.

bbannier:


bbannier:
Since this isn't exercised for PRs in this repo, yet, a sample run of this can be found [here](https://github.com/bbannier/btest/actions/runs/1142469526).

.github/workflows/btest.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  Run-BTest:
+    strategy:
+      matrix:
+        python_version: [3, 3.5]
+
+    runs-on: ubuntu-latest
+    container:
+      image: python:${{ matrix.python_version }}
```
ckreibich, line 13: If we switch this to Github's [suggested way](https://docs.github.com/en/actions/guides/building-and-testing-python) of running Python, we could also test on MacOS. Do you think it's worth it? It'd add a couple of lines to the YAML. I'm mainly wondering about the subtleties of the parallelization code, where I think we hit FreeBSD/MacOS vs Linux discrepancies a while back, no?

testing/Baseline/tests.sphinx.run-sphinx/output
```diff
@@ -1,23 +1,23 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 <div class="section" id="testing">
 <h1>Testing<a class="headerlink" href="#testing" title="Permalink to this headline">¶</a></h1>
-<div class="btest-cmd highlight-none"><table class="highlighttable"><tr><td class="linenos"><div class="linenodiv"><pre>1
-2</pre></div></td><td class="code"><div class="highlight"><pre><span></span><span class="hll"># echo Hello, world!
+<div class="btest-cmd highlight-none"><table class="highlighttable"><tr><td class="linenos"><div class="linenodiv"><pre><span class="normal">1</span>
```
ckreibich, line 4: There's a hiccup here with this baseline update: the tests will now break on systems that feature an older Sphinx (mine's on 3.4.3, fwiw). I suppose the format changes with 3.5? We could add a `TEST-REQUIRES` that looks for such a version.

rsmmr: Yeah, I've been seeing this diff for a little while locally too. Requiring a minimum Sphinx versions sounds right.

.github/workflows/btest.yml
bbannier: Yes, adding macos sounds worthwhile, so I changed this to use the documented approach. With that we'd have a Python version and a OS matrix, currently creating the following combinations:

- default python3 (seems to be 3.9)
    - ubuntu-latest
    - macos-latest
- python-3.5
    - ubuntu-latest
    - macos-latest

The macos setups always seem to ignore whatever `actions/setup-python@v2` does and in my tests both use the same version. I haven't look too deep into how to make it use another version (probably needs to be installed first), or how to ignore the redundant combination.

I also ran into inconsistencies in the expected output of `tests.sphinx.run-sphinx` between macos and ubuntu; even when forcing the same version of the `sphinx` module and its deps they seem to generate different HTML output (I suspect this pulls something from the system). I changed that test to now compare man output which is less noisy, but should still allow us to see our additional output.

testing/Baseline/tests.sphinx.run-sphinx/output
bbannier: I added an additional `TEST-REQUIRES` for the exact version of sphinx.

Since I still saw differences of the HTML output between macos and ubuntu, even with the same version of sphinx, I changed this test to now compare man output which has less noise.

bbannier:
Thanks for the review @ckreibich. An example run of the fixed setup can be found [here](https://github.com/bbannier/btest/runs/3410412997).

ckreibich (approved):
Looking great, thanks!

PR #54, opened 2021-08-25
Do not overwrite statefile on `KeyboardInterrupt`.

bbannier:
We would previously update the statefile when receiving a
`KeyboardInterrupt` during test execution and also update the test
stats even though this hardly ever makes sense in that scenario.

We now translate `KeyboardInterrupt` into an `Abort` which aborts any
further processing.


bbannier:
Closing this as this seems to cause other issues elsewhere.

PR #56, opened 2021-08-30
Switch Sphinx build test to use "text" output format

ckreibich:
We've hit a problem with our new CI runs because the Sphinx test puts the current date into the `man`-format output. This switches to `text`-format output, which seems to be the least sensitive to specific Sphinx versions. I've tried Sphinx versions >= 3.1.0 and they all pass.


ckreibich:
I found a few more problems while trying to reproduce the earlier CI failure on FreeBSD that are hopefully now fixed. The biggie is the locking of `Test.parseProgress()` between the test object and its monitoring thread. I think what happened here was they could execute in "parallel", triggering progress output to be reported twice. This seems to happen a ton on FreeBSD, but I could trigger it pretty rarely also in Linux.

ckreibich:
For reference, the earlier workflow I'm referring to is here: https://github.com/zeek/btest/runs/3467139295?check_suite_focus=true

testing/Baseline/tests.sphinx.run-sphinx/_build.man.btest-sphinxdemo.1
```diff
@@ -1,171 +0,0 @@
-### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-.\" Man page generated from reStructuredText.
-.
-.TH "BTEST-SPHINXDEMO" "1" "Aug 24, 2021" "1.0" "BTest-Sphinx Demo"
```
bbannier, line 4: Yikes, this doesn't work as a baseline at all (but 100% passed while looking at it). I like the `text` output even better than sanitized `man`.

bbannier (approved):

PR #59, opened 2021-09-16
Add basic pre-commit config.

bbannier:
The config for now only checks (and provides if run locally) basic
formatting, and performs linting of shell scripts. We fix any flagged
issues.


rsmmr:
I'll take a look at this.

btest-bg-wait
```diff
@@ -103,7 +109,7 @@ while true; do
         break
     fi
 
-    timeout=`expr $timeout - 1`
+    timeout=$((timeout - 1))
```
rsmmr: This is still a bashism, right?

btest-bg-wait
```diff
@@ -131,16 +137,16 @@ collect_output
 # See if any returned an error.
 result=0
 for p in $procs; do
-    if [ -e $p/.timeout ]; then
+    if [ -e "$p"/.timeout ]; then
```
rsmmr: -> `"$p/.timeout"`

rsmmr: There are more instance like this, would prefer to change them all.

sphinx/btest-rst-cmd
```diff
@@ -93,53 +93,57 @@ fi
...
+} >>"$stdout"
 fi
 
 for i in $include; do
-    echo "    `basename $i`" >>$filter_out
-    echo "" >>$filter_out
-    cat $i | apply_filter | indent >$filter_out
+    echo "    $(basename "$i")" >>"$filter_out"
+    echo "" >>"$filter_out"
+    apply_filter < "$i" | indent >"$filter_out"
```
rsmmr: Alright, this is hitting one of my pet peeves. :-) I find `cat xxx | foo | ...` much more readable than `foo <xxx | ...`

Can we turn this check off?

rsmmr: ... and we have more instances of this.

btest-ask-update
```diff
@@ -9,24 +9,33 @@
 # 200: No, don't update and abort.
 
 while true; do
-    printf "\033[0K" >>/dev/tty # Delete any augmented output.
-    echo " failed" >>/dev/tty
-    echo ">> Type 'c' to continue, 'd' to see diagnostics, 'u' to update baseline, and 'a' to abort." >/dev/tty
+	printf "\033[0K" >>/dev/tty # Delete any augmented output.
```
rsmmr, line 12: Let's not change indent level, that's just a style-preference thing.

btest
```diff
@@ -404,7 +428,8 @@ class TestManager(multiprocessing.managers.SyncManager):
                 if not t:
                     continue
 
-                if t.serialize and hash(t.serialize) % Options.threads != thread_num:
+                if t.serialize and hash(
+                        t.serialize) % Options.threads != thread_num:
```
rsmmr, line 432: Let's increase the line width, this is breaking to quickly.

btest
```diff
@@ -198,6 +211,7 @@ def replaceEnvs(s):
 
     return RE_ENV.sub(replace_with_env, s)
 
+
```
rsmmr, line 214: please disable this check, not sure why one would want 2 empty lines all the time.

btest
```diff
@@ -741,7 +773,9 @@ class Test(object):
         for (tag, regexp, multiple, optional, group1, group2) in Commands:
             if not optional and tag not in cmds:
                 if tag == "exec":
-                    error("%s: mandatory keyword '@TEST-EXEC' or '@TEST-EXEC-FAIL' is missing." % file)
+                    error(
```
rsmmr, line 776: Can we avoid breaking after the opening `(`?

rsmmr (commented):
So I'm generally on board running these tools, but they are pretty liberal at changing things right now. I left some comments on the bash changes, which aren't too bad (note that I often commented only on one instance, while they are more). But I didn't look through all the Python formatting changes, it's quite noisy. Did you try setting the options to keep the changes minimal? If not, I suggest trying that.

btest
rsmmr: Also, while we are at it: I'd actually like to remove our dependency on bash. We could run on a linter for that, too.

btest-bg-wait
bbannier: This file is currently a Bash script, but for me the construct also seems to work well in sh.

sphinx/btest-rst-cmd
bbannier: Hmm, this pattern even has its [own award](https://porkmail.org/era/unix/award.html) 😅. I'll look into this one after fixing the other issues you raised.

btest-ask-update
bbannier: We now use tabs for indention, is that not something we like? If you'd prefer that I can role this back and apply an indention of 4 spaces everywhere.

rsmmr: Oh, I see. Yeah, no tabs please, not used in this code base.

sphinx/btest-rst-cmd
rsmmr: I know. :-)

btest
bbannier: This is actually pretty much convention in the Python ecosystem due to PEP8, see [here](https://pep8.org/#blank-lines) so I'd hesitate to deviate from that, but there's a configuration flag for it if we really want it.

btest-ask-update
bbannier: Since I already needed to add a custom configuration, I also activated a couple of other flags which keep the diff to the existing style smaller.

btest-diff
```diff
@@ -144,7 +145,7 @@ fi
 
 tmpfiles=""
 delete_tmps() {
-    rm -f $tmpfiles 2>/dev/null
+    rm -f "$tmpfiles" 2>/dev/null
```
rsmmr, line 153: This isn't correct I believe: it's changing from deleting potentially multiple files to only one (which then won't exist).

btest-diff
```diff
@@ -77,7 +79,8 @@ is_canon_baseline() {
 
 # Prints the requested baseline to standard out if it is canonicalized
 # or we're using binary mode. Otherwise fails and prints nothing.
-get_baseline() {
```
rsmmr, line 79: Is there an option to leave the opening brace in the same line?

btest
```diff
@@ -2462,26 +2630,32 @@ RE_END_FILE = re.compile(CommandPrefix + "END-FILE")
...
+RE_PORT = ("port", re.compile(CommandPrefix + "PORT: *(.*)"), True, True, 1, -1)
+RE_INCLUDE_ALTERNATIVE = ("alternative", re.compile(CommandPrefix + "ALTERNATIVE: *(.*)"), True,
+                          True, 1, -1)
+RE_IGNORE_ALTERNATIVE = ("not-alternative", re.compile(CommandPrefix + "NOT-ALTERNATIVE: *(.*)"),
+                         True, True, 1, -1)
+RE_COPY_FILE = ("copy-file", re.compile(CommandPrefix + "COPY-FILE: *(.*)"), True, True, 1, -1)
+RE_KNOWN_FAILURE = ("known-failure", re.compile(CommandPrefix + "KNOWN-FAILURE"), False, True, -1,
+                    -1)
+RE_MEASURE_TIME = ("measure-time", re.compile(CommandPrefix + "MEASURE-TIME"), False, True, -1, -1)
+RE_DOC = ("doc", re.compile(CommandPrefix + "DOC: *(.*)"), True, True, 1, -1)
```
rsmmr, line 2646: This is strictly worse, I think. Is it possible to exclude a single block of code from re-formatting?

rsmmr (commented):
Ok, I'm (almost) good with the Python formatting now.

btest-diff
bbannier: Definitely, this is even due to explicit configuration I had done to keep the diff smaller, but we can move closer to the defaults.

btest-diff
bbannier: Ups, good catch. I'll add an exclusion here. I also went over all the other similar changes to make sure we didn't fall into the same trap elsewhere.

rsmmr (approved):
LGTM now, thanks, I'll merge.

PR #73, opened 2022-03-10
Make test `duplication-selection` independent of actual scheduling.

bbannier:
This test validates that only a single test is run exactly once, even if
requested multiple times. To validate that, we previously diffed the
full BTest log, but since this includes the worker ID it might produce
diffing failures even though the test should have passed.

With this patch we remove the output diffing and instead just validate
that the test is reported to succeed once.


testing/tests/duplicate-selection.test
```diff
@@ -1,5 +1,7 @@
+# %TEST-DOC: Validates that if the same test is requested multiple times, it still only runs once.
+#
 # %TEST-EXEC: btest -j t t t >>out 2>&1
-# %TEST-EXEC: btest-diff out
+# %TEST-EXEC: test $(grep -c 't ... ok' out) = 1
```
rsmmr, line 4:
```suggestion
# %TEST-EXEC: test $(grep -c 't ... ok' out) = 1
# %TEST-EXEC: test $(grep -c 't \.\.\. ok' out) = 1
```

rsmmr: I'll make that change when merging.

rsmmr (approved):

PR #76, opened 2022-11-14
Fix failing CI for latest python3

bbannier:


btest
```diff
@@ -682,6 +688,7 @@ class CmdLine:
     These commands can be provided by @TEST-{EXEC,REQUIRES} instructions, an
     Initializer, Finalizer, or Teardown, or their part-specific equivalents.
     """
+
```
rsmmr, line 685: Looks like all these new empty lines is a new change coming with the bump? I don't like it. :-)  Can we disable that?

.pre-commit-config.yaml
```diff
@@ -26,3 +26,16 @@ repos:
...
+  - id: vermin
+    name: Check minimal required Python version
+    # Minimal required Python version according to
+    # https://docs.zeek.org/en/master/install.html#required-dependencies.
+    # The value be kept consistent with `py-version` in `.pylintrc`.
+    entry: vermin -vv --target=3.5- .
+    language: python
+    pass_filenames: false
+    additional_dependencies:
+      - 'vermin'
```
rsmmr, line 41: I don't follow what this is doing. Why do we need to check the Python version in a pre-commit hook?

btest
bbannier: This came from the bump of yapf from 0.31.0 to 0.32.0, probably this PEP8-related change, https://github.com/google/yapf/commit/5fda04e1cdf50f548e121173337e07cc5304b752. I'll take out the bump.

.pre-commit-config.yaml
bbannier: This is a custom pre-commit check invoking [vermin](https://pypi.org/project/vermin/). I added this since at least I have to run this locally to make sure I do not accidentally introduce language features which are unsupported in the minimal Python version required by Zeek (python-3.5, which is now EOL).

I can take this out if you think this is not useful.

btest
bbannier: Actually, there's a config toggle for that, I'll instead adjust our config so that change is not made.

.pre-commit-config.yaml
bbannier: Dropped.

rsmmr (approved):

PR #77, opened 2022-11-16
Always use UTF-8 encoding for BTest input and output files.

bbannier:
We previously would use the system locale to determine what encoding to expect for BTest input files, either implicitly by not specifying an encoding, or explicitly by querying the current locale. To make matters worse, we recently migrated from `locale.getdefaultlocale` to `locale.getlocale` which behaves differently, mainly since `getdefaultlocale` behaved surprisingly was reason for it becoming deprecated with 3.11, see https://docs.python.org/3/library/locale.html; our change here essentially caused a regression if e.g., a `btest.cfg` contained environment variables modifying the locale.

This can now cause BTest to behave differently depending on whether environment variables like `LC_ALL` or similar are set, even when the intention was to only specify what locale to use for the programs under test.

With this patch we always use UTF-8 for BTest input and output files to make its behavior more deterministic. Uses of an implicit encoding where identified by running `python3 -X warn_default_encoding btest ...`. This adjusted behavior seems to capture what we and probably also users already seem to have expected.

### DIFF
--- a/btest
+++ b/btest
@@ -103,10 +103,7 @@ def platform():
 
 
 def getDefaultBtestEncoding():
-    if locale.getlocale() is None:
-        return 'utf-8'
-
-    return locale.getpreferredencoding()
+    return 'UTF-8'
 
 
 def validate_version_requirement(required: str, present: str):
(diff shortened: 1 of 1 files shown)


bbannier:
This is a regression fix for behavior introduced in #76 which I noticed when running the Spicy test suite contains locale-related settings in its `btest.cfg`, https://github.com/zeek/spicy/blob/765924e34f1ecf939d240d9c52f928f4656ea489/tests/btest.cfg#L31-L32.

PR #78, opened 2022-12-06
Exit with error when using `-a`/`--alternative` with an undefined alternative, and fix use of `-` shorthand in combinations

### LINKED ISSUE #58: btest -a doesn't check for whether alternative exists
`btest -a nosuchalternative` runs happily, reporting it's using `[nosuchalternative]`, even when that alternative doesn't exist.

ckreibich:
Resolves #58.


rsmmr (approved):

.github/workflows/btest.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   Run-BTest:
     strategy:
       matrix:
-        python-version: [3, 3.5]
+        python-version: [3]
```
bbannier, line 12: python-3.5 is still the minimal Python version required by Zeek, and we should check that BTest actually runs on it. I opened #81 to add back that check.

Issue #11, opened 2019-03-13
%DIR set incorrectly after @TEST-START-NEXT


ckreibich:
I think there might be a problem with the `%DIR` value: when using it in `@TEST-EXEC` statements the value seems to differ in test executions after a `@TEST-START-NEXT`. Since the source directory for such tests is the same, this shouldn't be happening, right?

I'm attaching a demo, see [example.zip](https://github.com/zeek/btest/files/2959307/example.zip)
```
$ unzip example.zip
Archive:  example.zip
   creating: example/
   creating: example/tests/
  inflating: example/tests/test.sh
  inflating: example/btest.cfg

$ cd example
$ cat tests/test.sh
# @TEST-EXEC: echo %DIR
# @TEST-START-NEXT

$ btest -A -t
[  0%] tests.test ... ok
[ 50%] tests.test-2 ... ok
all 2 tests successful

$ cat .tmp/tests.test/.stdout
/home/christian/tmp/example/tests

$ cat .tmp/tests.test-2/.stdout
/home/christian/tmp/example
```
It looks like in the tests following the first one, `%DIR` is just `%(testbase)s`, but I didn't dig in. I did verify though that at least in Zeek we currently have no btests using both `%DIR` and `@TEST-START-NEXT`....

jsiwek:
My expectation would be that it uses the same dir value just as you expected, too.  So seems like a bug.

PR #85, opened 2023-01-25
Better protection against garbled output with parallelized tests

### LINKED ISSUE #70: Output for `-A` is garbled if used with `-j`
If one uses the `-A` option together with `-j` one ends up with garbled output, e.g.,

```.console
$ btest -A tests/start-*
[  0%] tests.start-file ... ok
[ 25%] tests.start-next ... ok
[ 50%] tests.start-next-dir ... ok
[ 75%] tests.start-next-naming ... ok
all 4 tests successful

$ btest -A tests/start-* -j
[  0%] tests.start-file ... [  0%] tests.start-next ... [  0%] tests.start-next-dir ... [  0%] tests.start-next-naming ... ok
ok
ok
ok
all 4 tests successful```

### PR DESCRIPTION
**Pull request by ckreibich**
I've encountered two separate sources of garbled output with btest. The former happened only on its own testsuite when running without a TTY, looking like this:
```
$ cd testing
$ ../btest -j | cat
[#12] tests.alternatives-undefined ... ok
[#7] tests.alternatives-overwrite-env ... ok
[#9] tests.alternatives-reread-config ... ok
[#11] tests.alternatives-testbase ... ok
[#1] tests.abort-on-failure ... ok
[#8] tests.alternatives-reread-config-baselinedir ... ok
[#15] tests.basic-succeed ... ok
[#13] tests.basic-fail ... ok
                             [#2] tests.abort-on-failure-with-only-known-fails ... ok
                                                                                     [#2] tests.crlf-line-terminators ... ok
                                                                                                                            [#5] tests.alternatives-filter ... ok
                                                                                                                                                                 [#9] tests.canonifier ... ok
                                                                                                                                                                                             [#10] tests.alternatives-substitution ... ok
                                       [#11] tests.canonifier-cmdline ... ok
                                                                            [#8] tests.canonifier-fail ... ok
                                                                                                             [#3] tests.alternatives-baseline-dir ... ok
                                                                                                                                                        [#13] tests.copy-file ... ok
                                                                                                                                                                                    [#14] tests.baseline-dir-env ... ok
                     [#5] tests.diag-all ... ok
                                               [#9] tests.diag-file ... ok
                                                                          [#12] tests.brief ... ok
                                                                                                  [#11] tests.diff-brief ... ok
                                                                                                                               [#9] tests.environment-windows ... not available, skipped
                                                                                                                                                                                        [#6] tests.alternatives-keywords ... ok
                             [#3] tests.doc ... ok
[#16] tests.binary-mode ... ok
[#13] tests.duplicate-selection ... ok
[#15] tests.console ... ok
[#2] tests.diag ... ok
```
Turns out this was due to `console.test`, via a side-effect introduced by `script`.

The other source of garbling was as reported by Benjamin in #70, which this also addresses.

Resolves #70.

### DIFF
--- a/btest
+++ b/btest
@@ -1571,7 +1571,7 @@ class OutputHandler:
         """Called just before a test begins."""
 
     def testCommand(self, test, cmdline):
-        """Called just before a command line is exected for a trace."""
+        """Called just before a command line is executed for a trace."""
 
     def testProgress(self, test, msg):
         """Called when a test signals having made progress."""
@@ -1621,7 +1621,7 @@ class Forwarder(OutputHandler):
             h.testStart(test)
 
     def testCommand(self, test, cmdline):
-        """Called just before a command line is exected for a trace."""
+        """Called just before a command line is executed for a trace."""
         for h in self._handlers:
             h.testCommand(test, cmdline)
 
@@ -1663,6 +1663,12 @@ class Forwarder(OutputHandler):
 
 
 class Standard(OutputHandler):
+    """
+    The default output handler, writing plain lines with test outcome.
+
+    Each test result is reported. For parallelized operation, the output
+    includes the thread number processing the test.
+    """
     def testStart(self, test):
         self.output(test, self.threadPrefix(), nl=False)
         self.output(test, "%s ..." % test.displayName(), nl=False)
--- a/btest
+++ b/btest
@@ -1727,9 +1727,8 @@ class Console(OutputHandler):
     Normal = "\033[0m"
 
     def __init__(self, options):
-        OutputHandler.__init__(self, options)
+        OutputHandler.__init__(self, options, reopen_std_file(sys.__stdout__))
         self.show_all = True
-        self.stdout = reopen_std_file(sys.__stdout__)
 
     def testStart(self, test):
         msg = "[%3d%%] %s ..." % (test.mgr.percentage(), test.displayName())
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: btest [hunk 1]
```diff
@@ -1725,55 +1725,39 @@ class Console(OutputHandler):
 
     def __init__(self, options):
         OutputHandler.__init__(self, options)
-        self.show_all = True
-        self.stdout = reopen_std_file(sys.__stdout__)
```
awelzel, line 1723: Hmm, I'm staring at the OutputHandler base and that is using stderr if no file is provided to `_output()`. Is this commit  switching from using stdout to stderr for the Console output handler, or should I have stared longer?

https://github.com/zeek/btest/pull/85/commits/f0882b4845a50a2677eb22391f3192da2ce2d9ff#diff-e0eb63058e3b050717b3f3a68e3596a3363cb5ffd8b0f936a4ea65d316911bcbR1532-R1534
```
    def _output(self, msg, nl=True, file=None):
        if not file:
            file = reopen_std_file(sys.__stderr__)
```

I'm secondly somewhat wondering about all the "re-opening" that happens for each _output() call in that base class, but I guess it doesn't hurt as it's just wrapping.

If we actually switched to stderr, we probably should revert that to stdout. Keeping that `self.stdout` here and overwriting `.output()` in this class with something like the following might be enough?

```
def output(self, test, msg, nl=True, file=None):
   file = file or self.stdout  # or assert file is None (or ditch the file parameter altogether, but that may upset linters)
   return super().output(test, msg, nl=nl, file=self.stdout)
```

awelzel (changes requested):

btest
ckreibich: I can confirm that it switched destination, and that's unintentional. It should certainly remain identical for the console outputters, so let me fix that.

Re. the re-opening, Tim introduced that as part of the Windows support so I really didn't want to go near that. :-)

awelzel (approved):
:crossed_fingers:

PR #89, opened 2023-02-03
Fix monitor_cb thread preventing exit and output encoding issues

### LINKED ISSUE #88: Running -d with non-utf8 encoded characters in output hangs process
Non UTF-8 encoded byte sequences in the .diag / stderr output hang btest when run with `-d`.

We ran into this in the following Cirrus CI build, where the Alpine job hung due to Zeek outputting non-UTF8 encoded bytes on stderr.
https://cirrus-ci.com/task/6325848790269952?logs=test#L24

The below test triggers the hang on Debian and also Alpine. I'm a bit uncertain why the Zeek CI run above only hung on Alpine, but might have been a baseline difference only there (and the difference then being displayed).

There seem to be two things: We should probably ignore decoding errors by some `errors=` strategy. The fact that the process hangs upon and unhandled Exception is a separate scary thing.

```
$ cat x.test
@TEST-EXEC: bash -c 'echo -e "\xed\xee\xef" >&2 && false'

$ btest -d x.test 
[  0%] x ... failed
  % 'bash -c 'echo -e "\xed\xee\xef" >&2 && false'' failed unexpectedly (exit code 1)
  % cat .stderr
Process #1:
Traceback (most recent call last):
  File "/usr/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 500, in <lambda>
    name="#%d" % (i + 1), target=lambda: self.threadRun(i, mgr_data)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 578, in threadRun
    t.run(self)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 1337, in run
    run_cmdseq(seq)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 1285, in run_cmdseq
    need_teardown |= run_cmdseq(cmd)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 1307, in run_cmdseq
    self.mgr.testFailed(self)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 765, in testFailed
    self._output_handler.testFailed(test, msg)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 1716, in testFailed
    h.testFailed(test, msg)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 2058, in testFailed
    self.showDiag(test)
  File "/home/awelzel/corelight-oss/zeek/auxil/btest/btest", line 2032, in showDiag
    for line in open(f):
  File "/usr/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 0: invalid continuation byte

< ... hangs ... >
```

awelzel:
Should fix #88


awelzel:
This happened to @timwoj again when he backported a PR with a non-ascii/non-utf8 baseline into Zeek 5.0.

Is this PR too light on details, any other way to get more attention?

btest
```diff
@@ -2029,7 +2037,7 @@ class Diag(OutputHandler):
 
             if os.path.isfile(f):
                 self.output(test, "  % cat " + os.path.basename(f), True, self._file)
-                for line in open(f):
+                for line in open(f, encoding="utf-8", errors="backslashreplace"):
```
bbannier, line 2040: Nice, this is where this should be, not on when the fds were set up. Before this used a locale-dependent value for creating the output, but changing the used encoding here is probably fine and not a API-breaking change (this is e.g., unrelated to baselines).

testing/tests/binary-diag.test
```diff
@@ -0,0 +1,6 @@
+# %TEST-EXEC-FAIL: btest -d %INPUT 2>&1
+# %TEST-EXEC: btest-diff .stdout
+
+@TEST-EXEC: bash -c 'echo -e "Hey \xed\xee\xef!" >&2'
```
bbannier, line 4: I am confused, is this supposed to output non-ASCII characters? I see the literal escape sequences from the command in `.stdout`.

bbannier, line 4: I am confused, is this supposed to output non-ASCII characters? With the `bash` and its `echo` on macos I see unexpanded escape sequences in `.stdout`.

bbannier: I looked at the incorrect file. Due to nesting I would have needed to look at the `.tmp` folder inside the test's folder under `.tmp`. The data there looks like expected.

bbannier (approved):

testing/tests/binary-diag.test
awelzel: As in chat, sending it through `hexdump -C` and checking in `.tmp/tests.binary-diag/.tmp/binary-diag/.stderr` is were the original output exists. The one on the higher level is already btest-escaped.

PR #90, opened 2023-02-21
Support environment variable removal in alternatives

### LINKED ISSUE #60: Allow alternatives to unset an environment variable.
One can already override an existing environment variable (`FOO=bar`), but one cannot remove it; the best one can do is to set it to an empty value (`FOO=`)

Not sure on syntax, maybe: 
```
[environment-myalternative]
-FOO=
```

?

awelzel:
Closes #60.


awelzel:
Unfortunately, it doesn't seem the original motivation of unsetting `ZEEK_PROFILER_FILE` makes a difference anymore. Something might have been fixed here.

https://github.com/zeek/zeek/pull/1787

timwoj (approved):

timwoj:
> Unfortunately, it doesn't seem the original motivation of unsetting ZEEK_PROFILER_FILE makes a difference anymore. Something might have been fixed here.

I tried it here too and it doesn't seem to make any difference removing that flag. We must have fixed something else somewhere that sped it up.

PR #93, opened 2023-04-20
Alpine fixes

### LINKED ISSUE #92: `btest-diff` is broken on e.g., Alpine
Alpine ships with BusyBox `diff` which does not support the recently added option `--strip-trailing-cr`. This was working previously.

```console
$ cat .stderr
  diff: unrecognized option: strip-trailing-cr
  BusyBox v1.31.1 () multi-call binary.
  
  Usage: diff [-abBdiNqrTstw] [-L LABEL] [-S FILE] [-U LINES] FILE1 FILE2
  
  Compare files line by line and output the differences between them.
  This implementation supports unified diffs only.
  
  	-a	Treat all files as text
  	-b	Ignore changes in the amount of whitespace
  	-B	Ignore changes whose lines are all blank
  	-d	Try hard to find a smaller set of changes
  	-i	Ignore case differences
  	-L	Use LABEL instead of the filename in the unified header
  	-N	Treat absent files as empty
  	-q	Output only whether files differ
  	-r	Recurse
          --no-dereference Don't follow symlinks
  	-S	Start with FILE when comparing directories
  	-T	Make tabs line up by prefixing a tab when necessary
  	-s	Report when two files are the same
  	-t	Expand tabs to spaces in output
  	-U	Output LINES lines of context
  	-w	Ignore all whitespace
```

A possible workaround for users is to use the `diffutils` package.

ckreibich:
While working on #92 I noticed that two tests were broken when running on Alpine, so this fixes both. Would be good to get confirmation that the `btest-diff` tweak actually looks as expected on Windows, but I think that's covered by the testsuite.

Resolves #92.

### DIFF
--- a/btest-progress
+++ b/btest-progress
@@ -11,6 +11,20 @@ EOF
     exit 1
 }
 
+function get_date {
+    # Some date versions don't provide nanosecond substitution, yielding various
+    # strings instead ("." on Alpine, ".3NZ" on FreeBSD). Check if we got three
+    # digits and a "Z", otherwise fake the nanoseconds.
+    local res
+    res=$(date -u +'%Y-%m-%dT%H:%M:%S.%3NZ')
+
+    if echo "$res" | grep -q '[0-9][0-9][0-9]Z$'; then
+        echo "$res"
+    else
+        date -u +'%Y-%m-%dT%H:%M:%S.000Z'
+    fi
+}
+
 ### Main.
 
 quiet=0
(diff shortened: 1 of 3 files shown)


bbannier (approved):
LGTM modulo issues flagged by pre-commit, and assuming that the Windows CI actually tests this.

btest-progress
```diff
@@ -11,6 +11,16 @@ EOF
     exit 1
 }
 
+function get_date {
+    # Alpine's BusyBox date does not provide nanosecond granularity.
+    # Fill with zeros, so we still produce consistent output.
+    if date --help 2>&1 | grep -iq busybox; then
```
bbannier, line 17: Just as a note, instead of checking the output we could also just execute the full format string, then detect whether it ends in `.` and pad manually.

ckreibich: Thanks Benjamin — this had me look at this a bit more broadly and I just realized it doesn't work on FreeBSD either:
```
$ date -u +'%Y-%m-%dT%H:%M:%S.%3NZ'
2023-04-20T19:06:33.3NZ
```
For BusyBox I was surprised it doesn't end in `.Z`, and here's yet another variant. I'll work it over...

PR #100, opened 2024-01-08
Add option to read lists of tests from file.

rsmmr:
```
-i FILE, --tests-file=FILE
    Loads the list of tests to execute from a file. Each line in the
    file is interpreted as the name of a test, or a group of tests, to
    execute, just like the tests would be specified on the command
    line. Empty lines and lines starting with ``#`` are ignored. (This
    format is compatible with that of the ``btest`` `StateFile
    <state_file_>`_.)
```


bbannier (approved):
I think my points from #55 which we probably should link still stand, but I am not strongly opposed to this. Maybe but for the options conflict checking this is simple enough, though `btest $(cat <FILE>)` might have less limitations (e.g., can both "run from file" and specify additional tests on the command line).

btest
```diff
@@ -193,12 +193,12 @@ def getOption(key, default, section="btest"):
...
 
 
-def readStateFile():
+def readTestsFile(file):
     try:
         # Read state file.
         tests = []
 
-        for line in open(StateFile):
+        for line in open(file):
```
bbannier, line 201: Since the test file now can come from users let's make sure to enforce the expected encoding.

```suggestion
        for line in open(file, encoding="utf-8"):
```

rsmmr:
> I think my points from #55 

Oh, I totally forgot about that one and the discussion there ---  just found myself wanting such an option once more. That said, I do prefer this over `$(cat <FILE>)`: easier to both type and remember. And this isn't tied to the state file, so that should avoid any confusion about updating.

PR #101, opened 2024-03-01
Standardize on ruff for python formatting/linting via pre-commits

timwoj:
This is part of a minor effort to standardize all of our python pre-commits on the same tooling, in this case [ruff](https://github.com/astral-sh/ruff).


bbannier (approved):
I am 💯% on board with this. Do you believe removing a linter needs additional buy-in? If not fine if you merge it.

PR #104, opened 2024-07-30
Cleanup of project configuration and linting

bbannier:


btest
```diff
@@ -3183,16 +3198,16 @@ if __name__ == "__main__":
     )
 
     StateFile = normalize_path(
-        getOption("StateFile", os.path.join(defaults["testbase"], ".btest.failed.dat"))
+        getOption("StateFile", os.path.join(defaults["testbase"], ".btest.failed.dat")),
```
awelzel, line 3201: I find these extra commas very unusual, at least for function calls, but maybe that's just what it looks like everywhere today. I'd certainly vote for reverting "Active linter for style around trailing commas" if there was a choice. And add a magic trailing comma explicitly in places [where it looks better.](https://github.com/zeek/btest/pull/104/commits/23a9985fd7d9586c80e59b75f86bd8c671fafebb#diff-e0eb63058e3b050717b3f3a68e3596a3363cb5ffd8b0f936a4ea65d316911bcbR226)

The `%` to `f-string` change seems churny on first sight, too, but I guess it's nice that next time one doesn't need to resort to `%` for consistency.

bbannier: re: trailing commas, agreed that I probably wouldn't be that explicit when manually formatting, but it can help with diffs. I reverted changes for that. In that patch I had accidentally mixed in fixes linting for https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4 which is now its own commit.

> The % to f-string change seems churny on first sight, too, but I guess it's nice that next time one doesn't need to resort to % for consistency.

We were actually already trying to catch these with the previous pyupgrade linter, but that one didn't find nearly enough. Ruff finds a few more, but there are still some remaining (mostly ones with more involved expressions after `%`).

awelzel (approved):

PR #105, opened 2024-10-02
btest-bg-run-helper: Ignore SIGTERM to self during cleanup

awelzel:
The kill 0 invocation would kill the btest-bg-run-helper itself, causing the subsequent logic to not be executed, resulting in runaway processes on the system.

### DIFF
--- a/btest-bg-run-helper
+++ b/btest-bg-run-helper
@@ -3,8 +3,19 @@
 # Internal helper for btest-bg-run.
 
 cleanup() {
+    # Ignore SIGTERM during cleanup to prevent terminating
+    # this process when sending signals to the process group.
+    trap true SIGTERM
+
     if [ ! -e .exitcode ]; then
         echo 15 >.exitcode
+
+        # Send SIGTERM to all processes in the process group
+        # of the calling process.
+        #
+        # This should terminate any well-behaved background
+        # commands that were spawned by the program under test
+        # unless they started their own process group.
         kill 0 &>/dev/null
 
         if [ -n "$pid" ]; then
--- a/btest-setsid
+++ b/btest-setsid
@@ -5,8 +5,9 @@ import sys
 
 try:
     os.setsid()
-except Exception:
-    pass
+except Exception as e:
+    print(f"btest-setsid failed: {e!r}", file=sys.stderr)
+    exit(1)
 
 prog = sys.argv[1]
 args = sys.argv[1:]


awelzel:
I was tempted to add a `-SIGKILL` for the process group at the end of the cleanup, but IIUC in the worst case this may SIGKILL unrelated processes because `btest-setsid` ignores `setsid()` error situations.

btest-bg-run-helper
```diff
@@ -3,9 +3,21 @@
...
+    # this process when sending signals to the process group.
+    trap true SIGTERM
+
     if [ ! -e .exitcode ]; then
         echo 15 >.exitcode
-        kill 0 &>/dev/null
+
+        # Send SIGTERM to all processes of the process group.
+        # We do not use pid 0 here, as the btest-setsid helper
+        # may have failed and we don't want to send signals to
```
rsmmr, line 15: This sounds different than the commit message: "would kill the btest-bg-run-helper" vs the more restrictive "btest-setsid helper may have failed". Is there a specific situation where the latter would fail?

btest-bg-run-helper
```diff
@@ -3,9 +3,21 @@
...
+        # processes in a process group we never created.
+        #
+        # This will terminate any well-behaved background commands
+        # that were spawned by the program under test.
+        kill SIGTERM -${BASHPID} &>/dev/null
```
rsmmr, line 20: Shouldn't this be:

```suggestion
        kill -SIGTERM ${BASHPID} &>/dev/null
```

btest-bg-run-helper
awelzel: >  Is there a specific situation where the latter would fail?

I guess it should never really fail for us, but I was mislead by that exception handler:

https://github.com/zeek/btest/blob/8e8a51a71c0745488fdf996fe72edb987d26d8b3/btest-setsid#L6-L9


I'll re-spin, reverting to `kill 0`, but adding a print and exit 1 in the exception handler.

btest-bg-run-helper
```diff
@@ -3,8 +3,19 @@
...
+
     if [ ! -e .exitcode ]; then
         echo 15 >.exitcode
+
+        # Send SIGTERM to all processes in the process group
+        # of the calling process.
+        #
+        # This should terminate any well-behaved background
+        # commands that were spawned by the program under test
+        # unless they started their own process group.
```
rsmmr, line 18: This comment probably should have been there from the get-go ... :-)

rsmmr (approved):

PR #111, opened 2025-01-28
Require CMake 3.15, inline with other Zeek projects

timwoj:
I'm not sure why this wasn't changed when we changed all of the other ones way back when, but here we are.


awelzel (approved):
Not sure this is technically required, but suspect the only use is bundling with Zeek, so :+1:

timwoj:
> Not sure this is technically required, but suspect the only use is bundling with Zeek, so 👍

It isn't, other than silencing the warning from newer CMakes that they're dropping 3.5 support at some point in the near future.

PR #115, opened 2025-04-22
Look in parent directories for `btest.cfg`

evantypanski:
1) Makes btest consider more than just the current directory when looking for the config file - it will now look in parent directories until it finds it
2) Allows relative paths in the absence of a valid test
3) Allows running the tests from outside of the test suite (eg `btest zeek/testing/btest/language/while.zeek`)

~There are a couple edge cases I'll check tomorrow-or-so, therefore this is a draft. And the failing test is a little sad (where it finds the parent's btest config even though it should fail), may need a different config file name so that it doesn't clash with the test suite's btest config.~


btest
```diff
@@ -2708,6 +2712,23 @@ def outputDocumentation(tests, fmt):
...
+            return file_path
+
+        parent_dir = os.path.dirname(current_dir)
+
+        if parent_dir == current_dir:
+            return None
+
+        current_dir = parent_dir
+
+
```
awelzel, line 2731: Fly-by: Mind using `pathlib` for clarity?

```
def find_file(filename):
    current_dir = pathlib.Path(".").absolute()

    while current_dir.name:
        f = current_dir / filename
        if f.is_file():
            return str(f)

        current_dir = current_dir.parent

    return None
```

There's a test (`tests.btest-cfg` that takes really long and seems to spawn a lot of `btest` recursively for me). Probably, if `filename` is already absolute than we shouldn't do the recursion (and possibly even have a separate env variable to disable that behavior).

awelzel: > So how about by default it recurses out to parents, but if the config file isn't default, it doesn't look in parents?

Starting with that and ignoring the extra env sounds reasonable, actually, yeah.

evantypanski:
I updated it to make sure dots work from relative directories, but that specifies the test exactly. So the dot syntax should stay the same, where you can't specify a "relative" test with dots.

I think the edge cases are mostly covered too, dots seem to work the same and I think temp directories don't change since they `cd` to the test base anyway. It just changes 2 places files are found.

btest
```diff
@@ -2446,23 +2446,29 @@ class LinuxTimer(TimerBase):
 
 
 # Walk the given directory and return all test files.
-def findTests(paths, expand_globs=False):
+def findTests(paths, expand_globs=False, cwd=pathlib.Path.cwd()):
```
awelzel, line 2449: As in DM, I don't think we need to keep the API stable, so would also suggest to add a `*` after paths and adapt the callers, so  the places with True False are readable :)

testing/tests/resolve-from-cwd.test
```diff
@@ -0,0 +1,8 @@
+# %TEST-DOC: Ensure btest can find tests from a path relative to the current directory
+#
+# %TEST-EXEC: mkdir -p my/relative/dir
+# %TEST-EXEC: cat %INPUT >> my/relative/dir/a-relative-test.test
+# %TEST-EXEC: cd my/relative/dir && btest a-relative-test.test
+# %TEST-EXEC: cd my/relative && btest dir/a-relative-test.test
```
bbannier, line 6: It would be really cool if we could also handle cases where `btest` is run from a directory above or next to the folder with `btest.cfg` if a path to a test file is passed in, e.g., inside the root of a checkout of zeek/zeek:

```console
btest testing/btest/language/enum.zeek
```

This should be possible since we know that any `btest.cfg` should always be in the directory of the test or a parent.

If this is already handled let's add a test for it.

evantypanski: This raises the question for me - what would we expect for cases with multiple btest.cfg files?

```
.
├── btest.cfg
└── tests
    ├── sub1
    │   ├── Baseline
    │   ├── btest.cfg
    │   └── tests
    │       └── test2.test
    └── sub2
        └── test.test
```

Which `btest.cfg` does `test2.test` use? Do we care? Have we ever cared? Since there could be two strategies for choosing a config file:

1) Just go through the first file and its parents
2) Find the common subroot for all of the specified tests and check that directory and its parents

Seems simple implementation though, if that isn't an issue. I just don't really know

bbannier: My expectation would that the nearest one would be used, i.e., with the absolute path of the test repeatedly pop of the tail and check for `btest.cfg`; in this case for `test2.test` this should land on `tests/sub1/btest.cfg`, but on `./btest.cfg` for `test.test`.

evantypanski: I think if that's the case this becomes significantly more complicated, since there are a couple parts that seem to assume that one btest invocation == one set of options. It seems pretty doable but I'm reluctant to do it here to explode the PR

Though, that seems like a very good expansion IMO. I don't really like having to climb into the testing directory in order to execute btests..

bbannier: Sorry for being confusing, I think failing if both test files are requested is completely fine.

evantypanski: I took the plunge and did it in bf71a3e36dd6f54fc7b38fea742feb78082b6126 - if you have any thoughts on that I'm open. I found [`os.path.commonpath`](https://docs.python.org/3/library/os.path.html#os.path.commonpath) which was very welcome

It's kind of annoying that you can't do `btest -d testing/btest` in Zeek after this. But that's for the same reason you can't currently do `cd testing/btest && btest -d .` in Zeek - a different issue, just surfaced here. I think I've gone further than I initially wanted though

btest
```diff
@@ -2446,23 +2445,29 @@ class LinuxTimer(TimerBase):
...
 
     expanded = set()
 
-    for p in paths:
-        p = os.path.join(TestBase, p)
+    for unexpanded in paths:
+        expanded_path = os.path.join(TestBase, unexpanded)
+
+        # Allow relative directories if this one does not exist
+        if not os.path.exists(expanded_path):
```
awelzel, line 2460:
```suggestion
        if not os.path.exists(expanded_path) and cwd:
```

btest
```diff
@@ -2446,23 +2445,29 @@ class LinuxTimer(TimerBase):
...
     tests = []
 
     ignore_files = getOption("IgnoreFiles", "").split()
     ignore_dirs = getOption("IgnoreDirs", "").split()
 
     expanded = set()
 
-    for p in paths:
-        p = os.path.join(TestBase, p)
+    for unexpanded in paths:
```
awelzel, line 2456: I find unexpanded and expanded not super intuitive. Mostly because `glob.glob()` is doing the actual expansion IIUC. This is mostly about prefixing with testbase or cwd. Maybe keep the `p` and maybe "anchored" or "prefixed"

(nitpick)

btest
```diff
@@ -2991,18 +3016,34 @@ if __name__ == "__main__":
 
     (Options, args) = parse_options()
 
+    # Find the common path of the provided tests, if possible
+    try:
+        common_path = pathlib.Path(os.path.commonpath(args)).absolute()
```
awelzel, line 3021: I'm not sure that's really worth it. If one specifies two paths that end up with disagreeing `btest.cfg` files, seems just stopping is easier to understand than using the wrong one.

btest
```diff
@@ -2991,18 +3016,34 @@ if __name__ == "__main__":
...
+        common_path = pathlib.Path.cwd()
+
+    btest_cfg = find_config_file(
+        Options.config,
+        start_dir=common_path,
+        recurse=Options.config == DEFAULT_CONFIG_NAME,
+    )
+    if not btest_cfg:
+        error(f"configuration file '{Options.config}' not found")
+
```
awelzel, line 3034: WDYT about adding the following. Seem annoying that `btest testing/btest` woudln't work if we already do all this.

```
 
+    # Special case for providing just a single path to btest that happens
+    # to be the TestBase. Reset args, assuming the user wants to run all tests.
+    if len(args) == 1 and os.path.abspath(args[0]) == os.path.dirname(os.path.abspath(btest_cfg)):
+        args = []
+
```

btest
```diff
@@ -3066,6 +3107,7 @@ if __name__ == "__main__":
             transform=lambda x: normalize_path(x),
         )
 
+    cwd = pathlib.Path.cwd()
```
awelzel, line 3110:
```suggestion
    orig_cwd = pathlib.Path.cwd()
```

Nit (and not sure it's a good name), but might make sense to name it something that indicates that it's not the current cwd.

awelzel (approved):
Suggest to get another approval from @bbannier or @rsmmr.

Comments are mostly nitpickery, though it'd be nice to have the special case for the test base directory imo.

btest
evantypanski: Sorry I think I'm missing something, what's the proposed alternative? Searching from the first arg would silently work in a similar (but maybe) worse way. Not doing this if there are two+ tests seems heavy-handed. Then stopping would mean we somehow recognize the btest.cfg files conflict, which would mean going up from each arg to check?

Is there an alternative I'm missing? I am probably overcomplicating something here :)

awelzel: Let me resolve this. This is needed if one passes two individual tests I gather now.

The example from you commit message is the one that I was concerned, but in practice that probably won't happen all that much.

evantypanski: Yeah, my motivating example was multiple btest.cfg, but really there just needs some way to reconcile where to start searching when multiple tests are passed in. Determining that then determines which of the conflicting btest.cfg files gets selected. Then I find the common path to be the least surprising of the options for that selection

btest
```diff
@@ -2708,6 +2713,28 @@ def outputDocumentation(tests, fmt):
...
+        return None
+
+    current_dir = start_dir.absolute()
+
+    while current_dir.name:
+        f = current_dir / filename
+        if f.is_file():
+            return str(f)
+
+        current_dir = current_dir.parent
```
bbannier, line 2733: More as a suggestion, I think this could be written simpler like so:

```suggestion
    for d in [current_dir, *current_dir.parents]:
        f = d / filename
        if f.is_file():
            return str(f)
```

btest
```diff
@@ -3201,13 +3249,13 @@ if __name__ == "__main__":
 
     testdirs = getOption("TestDirs", "").split()
     if testdirs:
-        Config.configured_tests = findTests(testdirs, True)
+        Config.configured_tests = findTests(testdirs, cwd=None, expand_globs=True)
```
bbannier, line 3252: This is the caller we could simplify if `findTests` defaulted to `cwd=None`.

btest
```diff
@@ -3066,6 +3113,7 @@ if __name__ == "__main__":
             transform=lambda x: normalize_path(x),
         )
 
+    orig_cwd = pathlib.Path.cwd()
```
bbannier, line 3116: In all the twiddling in this patch I find the 100 lines between us storing this value and using it below the most painful. This is just due to how this code is structured, so nothing to do.

testing/tests/outside-btest-dir.test
```diff
@@ -0,0 +1,14 @@
...
+# %TEST-EXEC: mkdir -p base/tests/one
+# %TEST-EXEC: mkdir -p base/tests/two
+# %TEST-EXEC: cat %INPUT >> base/tests/one/a-relative-test.test
+# %TEST-EXEC: cat %INPUT >> base/tests/two/another-relative-test.test
+# %TEST-EXEC: mv btest.cfg base/
+# %TEST-EXEC: btest base/tests/one/a-relative-test.test
+# %TEST-EXEC: btest base/tests/one/a-relative-test.test base/tests/two/another-relative-test.test
+#
+# Test the special case of running all tests from outside
+# %TEST-EXEC: btest base/
```
bbannier, line 12: This will run zero tests since `btest.cfg` is actually in `base/tests/`. Let's fix that and maybe baseline the logs to make sure this does what we expect.

With your patch selecting tests by subdirectory also still, let's also test that here.

Maybe something like this?

```suggestion
# %TEST-EXEC: echo "## RUNNING DIRECTLY ON TEST DIR" >>output
# %TEST-EXEC: btest -v base/tests >>output 2>&1
# %TEST-EXEC: echo "## RUNNING DIRECTLY ON TEST SUBDIR" >>output
# %TEST-EXEC: btest -v base/tests/one >>output 2>&1
# %TEST-EXEC: btest-diff output
```

bbannier (approved):

btest
bbannier: Actually, using `current_dir.name` like originally fails if the config is at `/btest.cfg` and we attempt to climb up.

btest
```diff
@@ -2708,6 +2713,28 @@ def outputDocumentation(tests, fmt):
             print()
 
 
+# Finds the config file 'filename', recursing to parent directories from the
+# 'start_dir' if not found.
+def find_config_file(filename, *, start_dir, recurse):
+    # First check current dir
+    f = pathlib.Path(filename)
+    if f.is_file():
+        return str(f)
```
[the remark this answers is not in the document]
awelzel: :+1: 

Can then also remove the `abspath` from the "special case" of `os.path.dirname(os.path.abspath(btest_cfg))`.

testing/tests/outside-btest-dir.test
evantypanski: Hm the `btest.cfg` is actually in the `base/` directory, I went and confirmed. It actually just runs no tests because there is no `TestDirs` variable in that btest.cfg

I still think it's a good idea to baseline the output, since obviously I didn't catch that no tests ran and some should be ran, but that's (luckily) just a little weird issue

btest
```diff
@@ -2455,14 +2454,20 @@ def findTests(paths, expand_globs=False):
...
 
     for p in paths:
-        p = os.path.join(TestBase, p)
+        anchored = os.path.join(TestBase, p)
+
+        # Allow relative directories if this one does not exist
+        if not os.path.exists(anchored) and cwd:
+            from_cwd = cwd / p
+            if from_cwd.exists():
+                anchored = str(from_cwd)
```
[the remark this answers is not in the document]
evantypanski: Yeah doing that only if `not expand_globs` seems reasonable

btest
evantypanski: Normalize keeping around a root-level `/btest.cfg` for maximum chaos? :)

btest
evantypanski: Technically, the `chdir` call can go down about 4 lines of code. But I'll leave it

btest
bbannier: Something something containers?

btest
```diff
@@ -3010,20 +3016,35 @@ if __name__ == "__main__":
 
     (Options, args) = parse_options()
 
-    btest_cfg = find_file(Options.config, recurse=Options.config == DEFAULT_CONFIG_NAME)
+    # Find the common path of the provided tests, if possible
```
rsmmr, line 3019: Quoting the commit message:

> If you call btest with both test2.test and test1.test from this
> directory, the btest.cfg file should be predictable. The solution here
> was to use the "common path" of all specified tests, then look
> upwards from there. Therefore, if both tests are specified, they will
> use the top level btest.cfg.

Is using the top-level `btest.cfg` actually the right decision? Wouldn't `test1.test` require the lower-level config?

More generally, I have to say I find this logic pretty confusing. I get the overall use case, but do we really need that conflict resolving? Have you run into that in practice? Could we just bail out if we find multiple `best.cfg`  and not allow executing tests from outside the suite in that case?

btest
```diff
@@ -2708,6 +2713,29 @@ def outputDocumentation(tests, fmt):
...
+        return None
+
+    current_dir = start_dir.absolute()
+
+    for d in [current_dir, *current_dir.parents]:
+        f = d / filename
+        if f.is_file():
+            return str(f)
+
+        current_dir = current_dir.parent
```
rsmmr, line 2734:
```suggestion
```

I don't think this line is needed / does anything, right?

rsmmr (commented):
There are cases where we don't abort if not finding a `btest.cfg`:

```
# pwd
/Users/robin/work/zeek/master/auxil/btest/testing
# ../btest /etc
/etc/manpaths: mandatory keyword '@TEST-EXEC' or '@TEST-EXEC-FAIL' is missing..
```

One question: Once we have found `btest.cfg`, is the current working directory then consistently set to where it's located, independent from where `btest` was called? Is should, and from a quick test, I couldn't find a counter example; but from the code, I wasn't really sure.

btest
evantypanski: Sorry I made a mistake when applying Benjamin's suggestion

evantypanski:
>  is the current working directory then consistently set to where it's located, independent from where btest was called?

I believe so, it changes to the directory that the config file was found in. That's (thankfully) why all of the temp files and directories in the config just magically worked with this change. Seems like that's necessary for when a different config file is passed in, even before this change, so the machinery was there.

> There are cases where we don't abort if not finding a btest.cfg

Seems like btest also lets you execute ignored files as tests, it just adds them if it doesn't already know about them and you passed a file in. I personally would rather that case just error, but maybe it has some use?

btest
rsmmr: > I think the conflicts are rare enough that it doesn't matter too much unless it's consistent,

Ah, I see. I read the commit message as suggesting that you built it this way to *support* cases with multiple configs. But it sounds like instead its "multiple configs are so rare that it doesn't matter if they aren't working correctly". I agree that it's rare/almost impossible. That said, what you describe next is actually what I would find most intuitive:

> searching for each file's config file and comparing them against each other to see if they're the same.

Is that really that messy? Isn't it just one `find_config_file()` per path and seeing if the resulting joint set has exactly one element? But if it's not that simple, then I'm fine leaving as you have it.

rsmmr: > btest lets you execute ignored files if you pass in their path.

That's a bug I'd say.

rsmmr: > add an error if a test is in IgnoreFiles or IgnoreDirs.

Yeah, it's just pretty indirect way of catching the issue.

rsmmr (approved):
Ok, so approving, I don't want to hold this up on the rare/impossible case of multiple configs. If you can easily make that one change of finding all configs, I'd still prefer that, otherwise fine as is.

evantypanski:
For the test I had to update the `diff-remove-abspath` script (or it combined the two paths...). That was roughly from Zeek's (with modification).

This seems to be the consensus behavior, minus style nits

btest
```diff
@@ -2991,18 +3014,43 @@ if __name__ == "__main__":
 
     (Options, args) = parse_options()
 
+    found_configs = set()
+
+    for arg in args or [None]:
```
awelzel, line 3019: Would not make it that clever (short circuiting and `None` to denote cwd), but YMMV and doesn't matter really, feel free to ditch that comment.

```
-    for arg in args or [None]:
-        start_dir = pathlib.Path(arg).absolute() if arg else pathlib.Path.cwd()
+    if args:
+        for arg in args:
+            if found_config := find_config_file(
+                Options.config,
+                start_dir=pathlib.Path(arg).absolute(),
+                recurse=Options.config == DEFAULT_CONFIG_NAME,
+            ):
+                found_configs.add(found_config)
+    else:
         if found_config := find_config_file(
-            Options.config,
-            start_dir=start_dir,
-            recurse=Options.config == DEFAULT_CONFIG_NAME,
+            Options.config, start_dir=pathlib.Path.cwd(), recurse=True
         ):
             found_configs.add(found_config)
```

Or replacing the `[None]` trick with `[pathlib.Path.cwd()]` and dropping `start_dir = if .. else`.

btest
```diff
@@ -2991,18 +3014,43 @@ if __name__ == "__main__":
...
+            found_configs.add(found_config)
+
+    if len(found_configs) == 0:
+        error(f"configuration file '{Options.config}' not found")
+
+    if len(found_configs) > 1:
+        conflicting_configs = ", ".join(
+            sorted(f"'{config}'" for config in found_configs)
+        )
+        error(f"found multiple conflicting configuration files: {conflicting_configs}")
```
awelzel, line 3035: Total UX nit: Should maybe say something like: `cannot execute tests using to different configuration files together`.

awelzel (approved):

PR #116, opened 2025-04-25
Enable lints around return statements

bbannier:
This enables a class of return-related lints[^1] around return statements. I mainly came here for redundant returns after `else`[^2], but they all seemed useful. In general this seems to reduce nesting, and also make overall flow clearer.

[^1]: https://docs.astral.sh/ruff/rules/#flake8-return-ret
[^2]: https://docs.astral.sh/ruff/rules/superfluous-else-return/

### DIFF
--- a/btest
+++ b/btest
@@ -263,9 +263,7 @@ def ExpandBackticks(origvalue):
 
         return out.strip()
 
-    value = reBackticks.sub(_exec, origvalue)
-
-    return value
+    return reBackticks.sub(_exec, origvalue)
 
 
 # We monkey-patch the config parser to provide an alternative method that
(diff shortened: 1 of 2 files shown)


btest
```diff
@@ -1553,8 +1550,8 @@ class Test:
     def timePostfix(self):
         if self.utime_base >= 0 and self.utime >= 0:
             return f" ({self.utime_perc:+.1f}%)"
-        else:
-            return ""
+
+        return ""
```
evantypanski, line 1554: this one can also just be a ternary ¯\\\_(ツ)_/¯ not sure if I like it more though

evantypanski (approved):

btest
bbannier: And I thought I could just let the machines do the refactoring 😅.

With a ternary and our formatting this would like so

```diff
-        if self.utime_base >= 0 and self.utime >= 0:
-            return f" ({self.utime_perc:+.1f}%)"
-
-        return ""
+        return (
+            f" ({self.utime_perc:+.1f}%)"
+            if self.utime_base >= 0 and self.utime >= 0
+            else ""
+        )
```

While this is probably more compact I am not sure it buys us much in terms of readability. My hunch would be to keep as is.

PR #121, opened 2025-09-25
Check btest.cfg for Python configparser interpolation errors

### LINKED ISSUE #119: BTest [DEFAULT] variables only work in [btest] section, not others
The following (taken from the documentation [here](https://github.com/zeek/btest/blob/8c0fbfd74325b6c9be022a98bcd414b6f103d09e/README#L446-L465)):
```
[DEFAULT]
val=abcd

[btest]
ENV_VALUE1=%(val)s
```
works, but the following:
```
[DEFAULT]
val=abcd

[btest]
ENV_VALUE1=%(val)s

[environment]
ENV_VALUE2=%(val)s
```
generates a Python traceback, with the key line being: 
```
configparser.InterpolationSyntaxError: bad interpolation variable reference '%(val)'
```

ckreibich:
This adds a pass over the (non-default) config options in the freshly loaded configparser, catching any interpolation errors and reporting them in our own terms. This has some subtlety since we cannot just check all config options due to configparser's built-in default-fallback behavior. Since btest pulls os.environ into the defaults, checking everything may well trip over some environment variables that aren't formatted for Python interpolation (as, in my case, HISTTIMEFORMAT).

This is a follow-up to #119 because regardless of whether or not we find the default Python error message intuitive, I think crashing out with a backtrace is excessive.

Resolves #119.


ckreibich:
Btw — I am not sure whether we should still be patching `itemsNoDefaults` straight into the `ConfigParser` class. I can't find any reference of `ConfigParser.itemsNoDefaults` ever having been a method, so we're basically just plumbing a function into the class instead of inheriting, or simply defining a global function. I decided to steer clear of this here.

PR #136, opened 2026-06-01
Mark unstable tests as failures in JUnit output

timwoj:
The JUnit format only has tags for faiulre, error, and skipped. Our unofficial unstable tag causes problems with other tooling that reads the JUnit output, such as CircleCI's test insights.

This is related to https://github.com/zeek/zeek/issues/5512

### DIFF
--- a/btest
+++ b/btest
@@ -2169,7 +2169,6 @@ class XMLReport(OutputHandler):
     RESULT_PASS = "pass"
     RESULT_FAIL = "failure"
     RESULT_SKIP = "skipped"
-    RESULT_UNSTABLE = "unstable"
 
     def __init__(self, options, xmlfile):
         """Output handler for producing an XML report of test results.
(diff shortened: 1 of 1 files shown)


btest
```diff
@@ -2169,7 +2169,7 @@ class XMLReport(OutputHandler):
     RESULT_PASS = "pass"
     RESULT_FAIL = "failure"
     RESULT_SKIP = "skipped"
-    RESULT_UNSTABLE = "unstable"
+    RESULT_UNSTABLE = "failure"
```
bbannier: Works fine, but I'd prefer getting rid of this and then below updating `testUnstable`, e.g.,

```python
    def testUnstable(self, test, msg):
        # JUnit has no concept of flaky tests, report them as failures.
        self.testFailed(test, msg)
```

timwoj: That's reasonable 👍🏼

bbannier (approved):

awelzel:
Technically this is a breaking change and some users might have actually relied on this. I'm okay hand-wavy saying this will not impact anyone and it was against the spec, but I'm leaving a comment here anyhow.

There's no `NEWS` file, but it's in `CHANGES`, so that's hopefully enough.

timwoj:
True, but the documentation for btest also says this:

> -x FILE, --xml=FILE
>     Records test results in JUnit XML format to the given file.
>     If the file exists already, it is overwritten.

What we were outputting wasn't valid JUnit XML, and now it's more so.

PR #137, opened 2026-06-03
Automatically cancel exisiting GH CI workflows on PR update

timwoj:
Similar to https://github.com/zeek/zeek/pull/5531


bbannier (approved):
Like I mentioned offline, I suspect these particular builds are pretty fast and I doubt they are a big source of resource contention. Still not a bad idea to be conscious about what builds we care about.

PR #139, opened 2026-06-22
Prevent progress printing `100%` if not fully done

bbannier:
Previously we could print a test progress of `100%` even though there
were still tests running due to how we might round up, and for projects
with huge BTest suite this was pretty likely. With this patch we make
sure we _always_ round down to the nearest integer so we don't report
"all done" before we are actually done.


timwoj (approved):

timwoj:
I wondered about that, but not enough to go look at it. I'd love to figure out why zeek-testing/zeek-testing-private don't print out progress either like the rest of the btests too, but again not enough that I've prioritized it.

PR #140, opened 2026-07-23
Drop dependency on `multiprocess`

bbannier:
> [!WARNING]
> All code changes in this branch were generated with Claude Sonnet 4.6. I reviewed all changes and am accountable.

### DIFF
--- a/btest
+++ b/btest
@@ -39,8 +39,6 @@ DEFAULT_CONFIG_NAME = "btest.cfg"
 
 ConfigDefault = os.environ.get("BTEST_CFG", DEFAULT_CONFIG_NAME)
 
-# These regexes have fixed patterns and are used in child processes, so they
-# must be module-level constants to survive multiprocess spawn on Windows.
 RE_INPUT = re.compile(r"%INPUT")
 RE_DIR = re.compile(r"%DIR")
 RE_ENV = re.compile(r"\$\{(\w+)}")
@@ -446,8 +444,6 @@ class RunState:
             and self._failed > 0
             and self._failed > self._failed_expected
         ):
-            # Don't hand out any more tests if we are to abort after
-            # first failure. Doing so will let all the processes terminate.
             return None
 
         for i in range(len(self._tests)):
--- a/btest
+++ b/btest
@@ -350,51 +350,29 @@ def runTestCommandLine(cmdline, measure_time, **kwargs):
 # Runs a subprocess. Takes same arguments as subprocess.check_call()
 # and returns a 2-tuple (success, rc) where *success* is a boolean
 # indicating if the command executed, and *rc* is its exit code if it did.
-def _run_subprocess_mp(*args, **kwargs):
-    def child(q):
-        try:
-            if sys.platform == "win32":
-                tmpdir = normalize_path(kwargs.get("cwd", ""))
-                if len(args) > 1:
-                    cmd = shlex.join(args)
-                else:
-                    cmd = args[0]
-
-                tf, bash_cmd = _build_win_subprocess_cmd_script(cmd, tmpdir)
-                with tf:
-                    subprocess.check_call(bash_cmd, **kwargs)
-            else:
-                subprocess.check_call(*args, **kwargs)
-            success = True
-            rc = 0
-
-        except subprocess.CalledProcessError as e:
-            success = False
-            rc = e.returncode
-
-        except KeyboardInterrupt:
-            success = False
-            rc = 0
-
-        q.put([success, rc])
+def runSubprocess(*args, **kwargs):
+    if sys.platform == "win32":
+        tmpdir = normalize_path(kwargs.get("cwd", ""))
+        if len(args) > 1:
+            cmd = shlex.join(args)
+        else:
+            cmd = args[0]
+        tf, bash_cmd = _build_win_subprocess_cmd_script(cmd, tmpdir)
+        with tf:
+            proc = subprocess.Popen(bash_cmd, **kwargs)
+    else:
+        proc = subprocess.Popen(*args, **kwargs)
 
     try:
-        q = mp.Queue()
-        p = mp.Process(target=child, args=(q,))
-        p.start()
-        result = q.get()
-        p.join()
-
+        proc.wait()
     except KeyboardInterrupt:
-        # Bail out here directly as otherwise we'd get a bunch of errors.
-        # from all the childs.
-        sys.exit(1)
+        proc.kill()
+        proc.wait()
+        raise
 
-    return result
-
-
-def runSubprocess(*args, **kwargs):
-    return _run_subprocess_mp(*args, **kwargs)
+    if proc.returncode != 0:
+        return (False, proc.returncode)
+    return (True, 0)
 
 
 # Description of an alternative configuration.
--- a/btest
+++ b/btest
@@ -748,16 +748,6 @@ class TestManager(mp_managers.SyncManager):
     def run(self, tests, output_handler):
         self.start()
 
-        mgr_data = self.dict()
-        mgr_data["Alternatives"] = Alternatives
-        mgr_data["BaselineDirs"] = BaselineDirs
-        mgr_data["Initializer"] = Initializer
-        mgr_data["Finalizer"] = Finalizer
-        mgr_data["Teardown"] = Teardown
-        mgr_data["Options"] = Options
-        mgr_data["TestBase"] = TestBase
-        mgr_data["TmpDir"] = TmpDir
-
         lock = self.RLock()
         proxy_tests = self.list(tests)
         proxy_failed_tests = self.list([])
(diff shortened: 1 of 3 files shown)


btest
```diff
@@ -335,62 +307,84 @@ def replaceEnvs(s):
...
+    (success, rc) = await asyncRunSubprocess(
+        cmdline, cwd=cwd, env=env, stdout=stdout, stderr=stderr
+    )
     return (success, rc, -1)
 
 
 # Runs a subprocess. Takes same arguments as subprocess.check_call()
 # and returns a 2-tuple (success, rc) where *success* is a boolean
 # indicating if the command executed, and *rc* is its exit code if it did.
 def runSubprocess(*args, **kwargs):
```
awelzel: This can be dropped completely now. The only caller is Linux so can use `subprocess.call()` directly.

```
-# Runs a subprocess. Takes same arguments as subprocess.check_call()
-# and returns a 2-tuple (success, rc) where *success* is a boolean
-# indicating if the command executed, and *rc* is its exit code if it did.
-def runSubprocess(*args, **kwargs):
-    if sys.platform == "win32":
-        tmpdir = normalize_path(kwargs.get("cwd", ""))
-        if len(args) > 1:
-            cmd = shlex.join(args)
-        else:
-            cmd = args[0]
-        tf, bash_cmd = _build_win_subprocess_cmd_script(cmd, tmpdir)
-        with tf:
-            proc = subprocess.Popen(bash_cmd, **kwargs)
-            try:
-                proc.wait()
-            except KeyboardInterrupt:
-                proc.kill()
-                proc.wait()
-                raise
-        if proc.returncode != 0:
-            return (False, proc.returncode)
-        return (True, 0)
-    proc = subprocess.Popen(*args, **kwargs)
-
-    try:
-        proc.wait()
-    except KeyboardInterrupt:
-        proc.kill()
-        proc.wait()
-        raise
-
-    if proc.returncode != 0:
-        return (False, proc.returncode)
-    return (True, 0)
-
-
 async def asyncRunSubprocess(cmdline, *, cwd, env, stdout, stderr):
     if sys.platform == "win32":
         tmpdir = normalize_path(cwd)
@@ -2340,10 +2304,10 @@ class LinuxTimer(TimerBase):
             return False
 
         # Make sure it works.
-        (success, rc) = runSubprocess(
+        retcode = subprocess.call(
             f"{self.perf} stat -o /dev/null true 2>/dev/null", shell=True
         )
-        return success and rc == 0
+        return retcode == 0

```

btest
```diff
@@ -335,62 +307,84 @@ def replaceEnvs(s):
...
+        await proc.wait()
+    except asyncio.CancelledError:
+        proc.kill()
+        await proc.wait()
+        raise
 
-    return result
+    if proc.returncode != 0:
+        return (False, proc.returncode)
+    return (True, 0)
```
awelzel: The process awaiting is duplicatd in both paths, could separate into an inline helper. The less code specific to Windows, the better.

```
+    async def _result(proc):
+        """
+        Wait for process completion and build the return tuple.
+        """
+        try:
+            await proc.wait()
+        except asyncio.CancelledError:
+            proc.kill()
+            await proc.wait()
+            raise
+
+        if proc.returncode != 0:
+            return (False, proc.returncode)
+        return (True, 0)
+
-            try:
-                await proc.wait()
-            except asyncio.CancelledError:
-                proc.kill()
-                await proc.wait()
-                raise
-        if proc.returncode != 0:
-            return (False, proc.returncode)
-        return (True, 0)
+            return await _result(proc)
+
     proc = await asyncio.create_subprocess_shell(
         cmdline, cwd=cwd, env=env, stdout=stdout, stderr=stderr
     )
-
-    try:
-        await proc.wait()
-    except asyncio.CancelledError:
-        proc.kill()
-        await proc.wait()
-        raise
-
-    if proc.returncode != 0:
-        return (False, proc.returncode)
-    return (True, 0)
+    return await _result(proc)

```

btest
```diff
@@ -777,81 +605,160 @@ class TestManager(mp_managers.SyncManager):
...
+
+        out.close()
+
+
+async def run_test(test, state):
+    await test.run(state)
+    state.testReplayOutput(test)
+
+
+async def _async_run_all(state, n):
```
awelzel:
```suggestion
async def run_all_tests(state, n):
```

Don't think that needs a leading `_` here. The async is redundant also. `run_all_tests()` also fit the `run_test()` above more nicely.

btest
```diff
@@ -777,81 +605,160 @@ class TestManager(mp_managers.SyncManager):
...
+    async def worker(thread_num):
+        all_tests = []
+        while True:
+            thread_tests = state.nextTests(thread_num)
```
awelzel:
```suggestion
            tests = state.nextTests(thread_num)
```

Mostly because the "threads" in the name doesn't add a lot.

btest
```diff
@@ -1413,11 +1308,15 @@ class Test:
         self.stdout.close()
         self.stderr.close()
 
-        if self.monitor:
-            self.monitor_quit.set()
-            self.monitor.join()
+    async def _async_progress_monitor(self):
+        try:
+            while True:
+                self.parseProgress()
```
awelzel, line 1272: I didn't know there was a `.progress` feature.

btest
```diff
@@ -416,328 +410,163 @@ class Abort(Exception):
...
+            Options.abort_on_failure
+            and self._failed > 0
+            and self._failed > self._failed_expected
+        ):
+            return None
 
-            for i in range(len(self._tests)):
-                t = self._tests[i]
+        for i in range(len(self._tests)):
+            t = self._tests[i]
```
awelzel: Pre-existing, small modernization would be:
```
for i, t in enumerate(self._tests):
```

btest
awelzel: Indeed, the fact that `threadPrefix()` exists and thread and worker is mixed here, I find a bit unfortunate. Maybe `s/thread/worker/g` across the whole file? It's already a lot of changes and the consistency would be nice.

awelzel (commented):
This is clever. I like this.

I only looked at the final code though. Some minor nits and ideas while skimming the complete diff, but generally 👍

Drops the only dependency btest has, removes 168 lines of code, and running its own test-suite with -j4 gets a ~1.6x~ 1.7x speed-up in elapsed time for me locally (15.42/8.97 = 1.72, not sure how I got to 1.6x before...).

```
$ taskset -c 0,1,2,3 /usr/bin/time btest -j 4 -d testing/
91 tests successful, 3 skipped
22.99user 12.30system 0:15.42elapsed 228%CPU (0avgtext+0avgdata 62100maxresident)k
```

```
$ taskset -c 0,1,2,3 /usr/bin/time btest -j 4 -d testing/
91 tests successful, 3 skipped
26.54user 7.01system 0:08.97elapsed 373%CPU (0avgtext+0avgdata 61900maxresident)k
```

btest
```diff
@@ -1157,14 +1063,13 @@ class Test:
 
             time.sleep(15)
 
-    def run(self, mgr):
+    async def run(self, mgr):
```
awelzel:
```suggestion
    async def run(self, state):
```

btest
```diff
@@ -1157,14 +1063,13 @@ class Test:
...
         bound_sockets = self.getPorts(mgr, len(self.ports))
         self.bound_ports = [s.getsockname()[1] for s in bound_sockets]
 
         for bs in bound_sockets:
             bs.close()
 
-        self.progress_lock = threading.Lock()
         self.start = time.time()
         self.mgr = mgr
```
awelzel:
```suggestion
        self.state = state
```

That'd also be good because there's no manager anymore, so teh `self.mgr` / `mgr` is confusing. Everywhere else it seems to be called `state` now.

btest
bbannier: Done, also fixed other redundant `async` name prefixes in a cleanup.

btest
bbannier: Agree, mostly stayed away from that to avoid the churn, but then it would be inconsistent. I did a broader rename so things should be consistent again. I didn't touch existing CLI flags.

btest
bbannier: It is pretty niffty, but we went with default-off to avoid issues in terminals which didn't support it (e.g., my ghostty does, but the tmux I used doesn't).

awelzel: No, this is something else:

https://github.com/zeek/btest#displaying-progress

It's about calling `btest-progress` from within a test for propagation of progress information for a single test. I also now remember this is used at Corelight a bit for some tests, but I never realized what the implementation of this looks like, yikes.

bbannier: Oh my TIL also 🤦.

timwoj:
@bbannier Are you still planning on bringing this in? I'm a big fan. The multiprocess stuff really sucks to work with.

bbannier:
> Are you still planning on bringing this in? I'm a big fan. The multiprocess stuff really sucks to work with.

Yes, absolutely. I have been using exactly this state (well #141) locally without any issues, only just pushed an update which avoids a spurious error message; other than that this seems to work well. The only issue is that I haven't had found time to do a careful review of the code, so I have not been comfortable calling this "Ready for review". Let me try to do that today or tomorrow.

timwoj (approved):
I love it.

PR #141, opened 2026-07-27
Simplify how parallel workers are fed tests

bbannier:
We were using a pretty elaborate protocol with lists and `None` sentinel values. This PR simplifies so tests now read from a simple generator which encapsulates all the logic. We also do some adjacent cleanups.

### DIFF
--- a/btest
+++ b/btest
@@ -446,7 +446,7 @@ class RunState:
         for p in ports:
             self._ports.append(p)
 
-    def getAvailablePorts(self, count):
+    def available_ports(self, count):
         if count > len(self._ports):
             return []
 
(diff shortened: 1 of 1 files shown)


btest
```diff
@@ -430,18 +430,17 @@ class RunState:
...
+        while True:
+            for i, t in enumerate(self._tests):
+                if t.serialize and t.serialize_hash() % Options.threads != worker_num:
+                    # Not ours.
+                    continue
 
-        # No more tests for us.
-        return None
+                del self._tests[i]
+                yield from _expand_alternatives(t)
```
timwoj, line 440: This is interesting. Never seen that used before.

timwoj (approved):

PR #144, opened 2026-08-17
Error for `-u` without a TTY

evantypanski:
Agents often use `-u` then abort when they realize they can't actually interactively update changes, which sometimes caused processes to sit around. Instead, just abort early.

Change by me. Test by Claude.


awelzel (approved):
Seems fine. I've restarted CI as there seems to have been a spurious failure with one of the Windows tests.

PR #145, opened 2026-09-09
Add environment variable check to force console output

### LINKED ISSUE #143: Windows output displays differently than other platforms
Windows displays an output of the number of the concurrent job for each job vs a percent complete:

```
[#1] af_packet.plugin ... not available, skipped
[#2] bifs.addr_count_conversion ... ok
[#3] bifs.addr_to_ptr_name ... ok
[#4] bifs.addr_version ... ok
[#1] bifs.all_set ... ok
[#5] bifs.analyzer_name ... ok
[#6] bifs.anonymize_addr ... ok
[#7] bifs.anonymize_addr-2 ... ok
[#8] bifs.anonymize_addr-3 ... ok
[#3] bifs.any_set ... ok
[#4] bifs.backtrace ... ok
```

You can see that job #1 ends and then another job takes that slot. All of the other platforms display an overall percentage complete instead.

timwoj:
Taken from @awelzel's suggestion in https://github.com/zeek/btest/issues/143#issuecomment-5128587181.

Fixes #143

### DIFF
--- a/btest
+++ b/btest
@@ -2380,7 +2380,7 @@ def create_output_handler(options):
         output_handlers += [Brief(options)]
 
     else:
-        if sys.stdout.isatty():
+        if sys.stdout.isatty() or os.getenv("BTEST_FORCE_CONSOLE_OUTPUT", "0") == "1":
             if options.show_all:
                 output_handlers += [Console(options)]
             else:


btest
```diff
@@ -2380,7 +2380,7 @@ def create_output_handler(options):
         output_handlers += [Brief(options)]
 
     else:
-        if sys.stdout.isatty():
+        if sys.stdout.isatty() or os.getenv("BTEST_FORCE_CONSOLE_OUTPUT", "0") == "1":
```
bbannier, line 2383: WDYT about adding a proper option for this, and also reading a default from the environment like we do for `--progress-bar`?

https://github.com/zeek/btest/blob/1e97bbed58f8e56670115837b25f9317b0fc6773/btest#L3026-L3033

It would also be nice to add a test for this.

ckreibich: I agree — my suggestion would be to expand this PR a little bit. How about:
* add `--tty` to force console use
* use `BTEST_FORCE_TTY` as the corresponding variable
* wrap `sys.stdout.isatty() or os.getenv("BTEST_FORCE_TTY")` into a helper and use that instead of `sys.stdout.isatty()`. We have multiple of these and I'm not sure e.g. `--progress` works reliably in that Windows setting either.

If you have it in you it would also be handy to get a summary of supported `BTEST_*` env vars in the help output. I just had to remind myself and there are more than I would have come up with. :-)

